PR #10, opened 2024-01-30
feat: implement feature builder service

ldhyen99:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [x] 📕 storybook


packages/core/package.json
```diff
@@ -3,15 +3,16 @@
...
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "web-vitals": "^2.1.4"
   },
   "scripts": {
     "start": "craco start",
     "build": "craco build",
-    "test": "craco test",
-    "eject": "react-scripts eject"
+    "test": "jest --ci --env=jest-environment-jsdom --coverage"
```
huynguyen-hl, line 15: I think we shall add one more script for coverage.

packages/core/src/__tests__/GenericFeature.test.tsx
```diff
@@ -0,0 +1,120 @@
...
+          {name === 'Button' && <button onClick={props.onClick}>Click me!</button>}
+        </div>
+      ),
+    );
+
+    render(<GenericFeature components={componentsData} services={services} />);
+    fireEvent.click(screen.getByText('Click me!'));
+
+    expect(screen.getByText('Button')).not.toBeNull();
+    expect(services).toEqual([{ name: 'logService', parameters: [{ name: 'tester' }] }]);
```
huynguyen-hl, line 118: I'm a bit confused by this line because it asserts a variable object that is defined above to have its own data.

ldhyen99: Sorry, but I do not understand what you mean. Please explain more.

packages/core/src/components/GenericFeature/GenericFeature.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+  const executeServices = async (services: IServiceDefinition[], parameters: any[]) => {
+    return services.reduce(async (previousResult: any[] | Promise<any[]>, currentService) => {
+      const prevResult = await previousResult;
+      const service: any = getService(currentService.name);
+      const params = [...prevResult, ...currentService.parameters];
+      const result: any = service(...params);
+      return [result];
+    }, parameters);
+  };
```
huynguyen-hl, line 121: I don't see any function in this code block running asynchronously. Could you please explain why we use async/await here?

ldhyen99: This function executes the services with the accumulated parameters. We should put it asynchronously because it may have many services executed. For example, if we have two services, the second service should wait until the first service returns the value and receive this value as a param for the next service function.

packages/core/src/__tests__/GenericFeature.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+          {name === 'Button' && <button onClick={props.onClick}>Click me!</button>}
+        </div>
+      ),
+    );
+
+    render(<GenericFeature components={componentsData} services={services} />);
+    fireEvent.click(screen.getByText('Click me!'));
+
+    expect(screen.getByText('Button')).not.toBeNull();
```
namhoang1604, line 119: It should check the output after triggering the button

namhoang1604 (changes requested):

packages/services/src/types/index.ts
```diff
@@ -0,0 +1,3 @@
+export interface IServices {
+  getFormInfo(data: unknown): unknown;
+}
```
namhoang1604, line 3: I think it should be generic interface
```
interface IService {
  (arg1: any, ...args: any[]): any;
}
```

namhoang1604 (approved):

PR #13, opened 2024-02-20
feat: add scanning and verify pages

huynguyen-hl:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [x] 📕 storybook


packages/core/src/components/JsonBlock/JsonBlock.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { Button, Card, CardContent } from '@mui/material';
+import { VerifiableCredential } from '@vckit/core-types';
+
+const JsonBlock = ({ credential }: { credential: VerifiableCredential }) => {
+  /**
+   * handle click on download button
+   */
+  const handleClickDownloadVC = async () => {
```
ldhyen99, line 9: Do you think we should write this in function for reusable?

packages/core/.babelrc.json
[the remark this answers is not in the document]
huynguyen-hl: I'm sorry to have confused you. It's just used for running Jest unit tests. I noticed the problem when I tried to answer you. I just pushed another commit to make the configuration in the core package cleaner. Thank you for the good question.

packages/core/src/components/JsonBlock/JsonBlock.tsx
huynguyen-hl: Yes, I agree with you, but currently, it's only used in this place. So I don't want to make the core package more complex by splitting this function to another place. When another component uses this function, then I will move it to the utils file to facilitate easier reuse. Thank you for your feedback.

namhoang1604 (approved):

ldhyen99 (approved):

PR #16, opened 2024-02-26
feat: add transaction event service

huynguyen-hl:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📕 storybook

### DIFF
--- a/packages/services/src/__tests__/mocks/constants.ts
+++ b/packages/services/src/__tests__/mocks/constants.ts
@@ -139,5 +139,7 @@ export const contextTransactionEvent = {
 export const dataTransactionEvent = {
   data: {
     livestockIds: ['9359502000010'],
+    sender: 'test-1234',
+    receiver: 'test-5678'
   },
 };
--- a/packages/services/src/__tests__/mocks/constants.ts
+++ b/packages/services/src/__tests__/mocks/constants.ts
@@ -112,34 +112,19 @@ export const dataObjectEvent = {
 
 /*============================= */
 
-export const contextTransactionEvent = {
-  vckit: {
-    vckitAPIUrl: 'https://vckit.example.com',
+export const transactionEventMock = {
+  nlisidMock: '9988776600000',
+  uploadedTransactionEventLinkMock: `https://s3.ap-southeast-2.amazonaws.com/9988776600000`,
+  transactionEventDLRMock: `https://example.com/nlisid/9988776600000?linkType=all`,
+  transactionVCMock: {
+    '@context': ['https://example.sh/TransactionEvent.jsonld'],
+    type: ['VerifiableCredential', 'TransactionEventCredential'],
     issuer: 'did:web:example.com',
+    credentialSubject: {
+      parentItem: { itemID: `https://example.com/nlisid/9988776600000?linkType=all`, name: 'Beef Steak Variety Container' },
+      childItems: [{ itemID: 'http://example.com/beef-scotch-box.json', name: 'Beef Scotch Fillet Box' }],
+      childQuantityList: [{ productClass: 'Beef', quantity: '50', uom: 'units' }],
+      locationId: 'https://plus.codes/123MHR+PW',
+    },
   },
-  dpp: {
-    context: ['https://www.w3.org/2018/credentials/v1'],
-    renderTemplate: [{ template: '<p>Render dpp template</p>', '@type': 'WebRenderingTemplate2022' }],
-    type: ['TransactionEventCredential'],
-    dlrLinkTitle: 'Livestock Passport',
-    dlrIdentificationKeyType: 'nlisid',
-    dlrVerificationPage: 'https://web.example.com/verify',
-  },
-  dlr: {
-    dlrAPIUrl: 'http://dlr.example.com',
-    dlrAPIKey: '1234',
-  },
-  storage: {
-    storageAPIUrl: 'https://storage.example.com',
-    bucket: 'test-verifiable-credentials',
-  },
-  identifierKeyPaths: ['livestockIds'],
-};
-
-export const dataTransactionEvent = {
-  data: {
-    livestockIds: ['9359502000010'],
-    sender: 'test-1234',
-    receiver: 'test-5678'
-  },
-};
+}
--- a/packages/services/src/__tests__/mocks/constants.ts
+++ b/packages/services/src/__tests__/mocks/constants.ts
@@ -133,13 +133,11 @@ export const contextTransactionEvent = {
     storageAPIUrl: 'https://storage.example.com',
     bucket: 'test-verifiable-credentials',
   },
-  identifierKeyPaths: ['herd', 'NLIS'],
+  identifierKeyPaths: ['livestockIds'],
 };
 
 export const dataTransactionEvent = {
   data: {
-    herd: {
-      NLIS: 'NH020188LEJ00012',
-    },
+    livestockIds: ['9359502000010'],
   },
 };
--- a/packages/services/src/__tests__/mocks/constants.ts
+++ b/packages/services/src/__tests__/mocks/constants.ts
@@ -114,17 +114,18 @@ export const dataObjectEvent = {
 
 export const transactionEventMock = {
   nlisidMock: '9988776600000',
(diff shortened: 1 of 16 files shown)


packages/core/src/constants/app-config.json
```diff
@@ -240,6 +240,36 @@
                   "identifierKeyPaths": ["herd", "NLIS"]
                 }
               ]
+            },
+            {
+              "name": "processTransactionEvent",
```
ldhyen99, line 245: I think you need to create a new feature that has components and services for testing this event. Because when you put this in here after the UI triggers the processObjectEvent function, it'll call the processTransactionEvent function, and receive params is the return of processObjectEvent, it is not correct as we expect. 

The UI does not need to be complicated, I recommend using the JSON form with a few fields as input.

packages/core/src/constants/app-config.json
```diff
@@ -243,6 +243,111 @@
...
+                  "dpp": {
+                    "context": ["https://dpp-json-ld.s3.ap-southeast-2.amazonaws.com/transaction-event-ld.json"],
+                    "renderTemplate": [
+                      { "template": "<p>Render dpp template</p>", "@type": "WebRenderingTemplate2022" }
+                    ],
+                    "type": ["TransactionEventCredential"],
+                    "dlrLinkTitle": "Livestock Passport",
+                    "dlrIdentificationKeyType": "nlisid",
+                    "dlrVerificationPage": "http://localhost:3000/verify"
+                  },
```
namhoang1604, line 337: The transaction event does not issue the dpp (digital product passport). It just issue the transaction event VC.

packages/services/src/epcisEvents/transactionEvent.ts
```diff
@@ -0,0 +1,103 @@
...
+  livestockIds: string[];
+}
+
+export const processTransactionEvent: IService = async (transactionEvent: ITransactionEvent, context: IContext): Promise<VerifiableCredential> => {
+  const validationResult = validateContextObjectEvent(context);
+  if (!validationResult.ok) {
+    throw new Error(validationResult.value);
+  }
+
+  const { sender, receiver, locationUrl, industryType, transaction, livestockIds = [] } = transactionEvent.data as ITransactionData;
```
namhoang1604, line 34: Should not hardcode `{ sender, receiver, locationUrl, industryType, transaction, livestockIds = [] }`

namhoang1604 (changes requested):

packages/services/src/__tests__/mocks/constants.ts
```diff
@@ -111,3 +111,35 @@ export const dataObjectEvent = {
 };
 
 /*============================= */
+
+export const contextTransactionEvent = {
```
ldhyen99, line 115: please delete this code if you do not call it

packages/services/src/__tests__/transactionEvent.test.ts
ldhyen99, line 1: I recommend adding some test cases that a few fields in the object doest not validate, for example: what happens if ` epcisTransactionEvent` in context is empty?

packages/services/src/epcisEvents/transactionEvent.ts
```diff
@@ -0,0 +1,69 @@
...
+import { ITransactionEventContext } from './types.js';
+import { getIdentifierByObjectKeyPaths } from './helpers.js';
+import { uploadJson } from '../storage.service.js';
+import { generateUUID } from '../utils/helpers.js';
+import { registerLinkResolver } from '../linkResolver.service.js';
+import { validateTransactionEventContext } from './validateContext.js';
+import { EPCISEventAction, EPCISEventDisposition, EPCISEventType } from '../types/epcis.js';
+
+export interface ITransactionEvent {
+  data: {
```
ldhyen99, line 13: Should we create a common interface to reusable?

packages/services/src/epcisEvents/transactionEvent.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  const { vckit, epcisTransactionEvent, dlr, storage, identifierKeyPaths } = context;
+  const identifier: string = getIdentifierByObjectKeyPaths(transactionEvent.data, identifierKeyPaths);
+  if (!identifier) {
+    throw new Error('Identifier not found');
+  }
+
+  const credentialSubject = {
+    ...transactionEvent.data,
```
ldhyen99, line 31: Should we build a common function for reusable in other events?

packages/services/src/__tests__/transactionEvent.test.ts
huynguyen-hl: Your idea seems like we need to test the side effect function. I have a test case titled "should throw an error if context validation throws an error" to simulate an error for the `validateTransactionEventContext` function. It covers the scenario where we have an empty context passed to the function.

packages/services/src/epcisEvents/transactionEvent.ts
huynguyen-hl: I don't think we need to separate functions for the credentialSubject because each event has very different credentialSubject data. If we build a common function, it can take a lot of parameters. So, for now, I just want to keep it simple for better readability and easier adjustment.

ldhyen99: Ah, I mean the fields in `credentialSubject` such as `eventID`, ` eventType`, etc. Because I see it appears many times.

packages/services/src/__tests__/mocks/constants.ts
```diff
@@ -111,3 +111,20 @@ export const dataObjectEvent = {
...
+  transactionEventDLRMock: `https://example.com/nlisid/9988776600000?linkType=all`,
+  transactionVCMock: {
+    '@context': ['https://example.sh/TransactionEvent.jsonld'],
+    type: ['VerifiableCredential', 'TransactionEventCredential'],
+    issuer: 'did:web:example.com',
+    credentialSubject: {
+      parentItem: { itemID: `https://example.com/nlisid/9988776600000?linkType=all`, name: 'Beef Steak Variety Container' },
+      childItems: [{ itemID: 'http://example.com/beef-scotch-box.json', name: 'Beef Scotch Fillet Box' }],
+      childQuantityList: [{ productClass: 'Beef', quantity: '50', uom: 'units' }],
+      locationId: 'https://plus.codes/123MHR+PW',
```
namhoang1604, line 127: The mock looks like the aggregation event, can you please check again?

packages/services/src/__tests__/transactionEvent.test.ts
```diff
@@ -0,0 +1,150 @@
...
+      jest.spyOn(helpers, 'getIdentifierByObjectKeyPaths').mockReturnValueOnce(nlisidMock);
+      jest.spyOn(publicAPI, 'post').mockRejectedValueOnce("Can't issue VC");
+
+      await processTransactionEvent(transactionEvent, context);
+    } catch (e) {
+      const error = e as Error;
+      expect(error.message).toBe("Can't issue VC");
+      expect(validateContext.validateTransactionEventContext).toHaveBeenCalled();
+      expect(helpers.getIdentifierByObjectKeyPaths).toHaveBeenCalled();
+    }
```
namhoang1604, line 108: the `processTransactionEvent` does not handle try-catch when error throw for the calling services

namhoang1604 (changes requested):

packages/services/src/__tests__/transactionEvent.test.ts
huynguyen-hl: I think we should keep the original error thrown by the `processTransactionEvent` function. The developer implementing that function will catch those errors and handle them at their discretion.

namhoang1604 (approved):

ldhyen99 (approved):

PR #23, opened 2024-03-06
feat: dynamic schema loading

ldhyen99:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📕 storybook


packages/untp-test-suite/src/core/services/dynamic-loading-schemas/utils.ts
```diff
@@ -0,0 +1,62 @@
...
+};
+
+/**
+ * Check if a schema exists
+ * @param schemaName - The name of the schema
+ * @returns A promise that resolves to a boolean indicating if the schema exists
+ */
+export const checkSchemaExists = async (schemaName: string): Promise<boolean> => {
+  const targetPath = goUpLevels(process.cwd(), 3);
+  return checkFileExists(`${targetPath}/schemas/${schemaName}`);
```
namhoang1604, line 39: The schemas folder path is never changed, so I think it should be a constant, you don't need to implement a function `goUpLevels`

packages/untp-test-suite/src/core/services/dynamic-loading-schemas/types.ts
```diff
@@ -0,0 +1,3 @@
+export interface IDynamicLoadingSchemaService {
+  (schema: string, version: string): unknown;
```
namhoang1604, line 2: the output should be json

namhoang1604 (changes requested):

namhoang1604 (approved):

packages/untp-test-suite/src/core/services/dynamic-loading-schemas/loadingSchema.service.ts
```diff
@@ -0,0 +1,26 @@
...
+    const content = await getSchemaContent(schema, version);
+    if (!content) {
+      throw new Error(`Content in ${schema} schema not found`);
+    }
+
+    return JSON.parse(content);
+  } catch (e) {
+    const error = e as Error;
+    throw new Error(error.message ?? 'Error loading schema');
+  }
```
huynguyen-hl, line 25: I don't think we should use a try-catch block here because the error is thrown in the try block, which will then be caught in the catch block. As you implemented, you throw it again in the catch block. Logically, it's no different.

packages/untp-test-suite/src/core/services/dynamic-loading-schemas/types.ts
```diff
@@ -0,0 +1,3 @@
+export interface IDynamicLoadingSchemaService {
+  (schema: string, version: string): Promise<JSON>;
+}
```
huynguyen-hl, line 3: The function that uses `IDynamicLoadingSchemaService` doesn't return a JSON type inside the Promise. Instead, it returns an object. I think placing the JSON type here may not be entirely correct.

packages/untp-test-suite/src/core/services/dynamic-loading-schemas/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Get the content of a schema
+ * @param folderName - The name of the schema
+ * @param version - The version of the schema
+ * @returns A promise that resolves to the content of the schema
+ */
+export const getSchemaContent = (folderName: string, version: string): Promise<string> => {
+  return fs.promises.readFile(`${SCHEMA_PATH}/${folderName}/${version}/schema.json`, 'utf-8');
+};
```
huynguyen-hl, line 48: I think when we use the `fs.readFile` function, it handles checking whether the file exists or not. If the file does not exist at the specified path, it will throw an error for us. This error is very clear, allowing the developer who is implementing to easily understand the issue. Therefore, we don't need to manually check whether every subfolder contains the `schema.json` file.

packages/untp-test-suite/src/core/services/index.ts
```diff
@@ -1 +1,2 @@
 export * from './json-schema/index.js';
+export * from './dynamic-loading-schemas';
```
huynguyen-hl, line 2: The `untp-test-suite` is a module package, so I think we need to include '.js' at the end of the file import line.

namhoang1604 (dismissed):

packages/untp-test-suite/src/core/services/dynamic-loading-schemas/utils.ts
namhoang1604: > I think when we use the `fs.readFile` function, it handles checking whether the file exists or not. If the file does not exist at the specified path, it will throw an error for us. This error is very clear, allowing the developer who is implementing to easily understand the issue. Therefore, we don't need to manually check whether every subfolder contains the `schema.json` file.

We want to wrap up the loading file for schemas to a separating service, then the structure is quite clear and easy to change in the future. The wrapper here is also returning the specific message to easy understand

huynguyen-hl (approved):

namhoang1604 (approved):

PR #25, opened 2024-03-08
feat: Integrate interface with core runner function

huynguyen-hl:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [x] 📜 README.md
- [ ] 📕 storybook


packages/untp-test-suite/src/core/process-test-suite.ts
```diff
@@ -1,22 +1,22 @@
...
+import { readConfigFile, readFile } from './utils/common.js';
+import { dynamicLoadingSchemaService } from './services/dynamic-loading-schemas/loadingSchema.service.js';
+import { hasErrors } from './services/json-schema/validator.service.js';
 
-export const processTestSuite: TestSuite = async (credentialConfigsPath) => {
-  const credentialConfigs = await readFile<ConfigContent[]>(credentialConfigsPath);
-  validateCredentialConfigs(credentialConfigs);
-
-  const testSuiteResultPromises = credentialConfigs.map<Promise<TestSuiteResult>>(async (credentialConfig) => {
+export const processTestSuite: TestSuite = async () => {
```
huynguyen-hl, line 7: I think this function should have a parameter to specify the path of the credential configuration file we want to read. This would make the function more dynamic and easier to reuse. For example, in the `untp test` command, which calls this function, there could be an option for the user to specify the path of the credential configuration file (--config \<path\>) they want to use when running the test suite.

packages/untp-test-suite/src/core/process-test-suite.ts
```diff
@@ -1,22 +1,22 @@
...
+    ]);
+
     const errors = hasErrors(schema, data);
-    return {
-      ...credentialConfig,
-      result: errors ? 'FAIL' : 'PASS',
-      errors: errors ? ajv.errorsText(errors) : null,
-    };
+
+    return errors;
```
huynguyen-hl: I think we need to return the credential config along with the errors. Then, where this function is called, it will know which credential config has errors and which ones don't. 
For example, we can add the `errors` field in the return statement like this:

```ts
return {
  ...credentialConfig,
  errors
};
```

packages/untp-test-suite/src/core/utils/common.ts
```diff
@@ -29,6 +15,13 @@ export const readFile = async <T>(filePath: string): Promise<T> => {
   return JSON.parse(fileContent);
 };
 
+export const readConfigFile = async () => {
+  const configPath = path.resolve(process.cwd(), '../../config/credentials.json');
+  const result = await readFile<ConfigCredentials>(configPath);
+  validateCredentialConfigs(result.credentials);
```
huynguyen-hl, line 21: To make the `readConfigFile` function easier to understand, I suggest it should only do what its name implies: read the configuration file. The `validateCredentialConfigs` should be called where we call the `readConfigFile` function. This approach follows the principle of clean code, which emphasizes that each function should have a single, well-defined purpose.

packages/untp-test-suite/src/core/utils/common.ts
```diff
@@ -29,6 +15,13 @@ export const readFile = async <T>(filePath: string): Promise<T> => {
   return JSON.parse(fileContent);
 };
 
+export const readConfigFile = async () => {
```
huynguyen-hl, line 18: If we add a parameter to specify the path of the credential configuration file we want to read, I think it should be passed to this function. Since this function is primarily responsible for reading the credential configuration file.

namhoang1604 (approved):

ldhyen99 (approved):

PR #27, opened 2024-03-14
feat: add untp test suite cli tool

huynguyen-hl:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📕 storybook


namhoang1604 (approved):

packages/untp-test-suite/src/interfaces/utils/common.ts
```diff
@@ -0,0 +1,12 @@
+import { readFile } from 'fs/promises';
+
+export async function readJsonFile<T>(filePath: string): Promise<T | null> {
+  try {
+    const fileJson = await readFile(filePath, { encoding: 'utf-8' });
+    const file = JSON.parse(fileJson);
+
+    return file;
+  } catch (error) {
+    return null;
```
ldhyen99, line 10: I am confused about return null in catch. Will we need this error in some place?

huynguyen-hl: Sometimes, we need to throw the error to the outside where this function is called, and sometimes we don't. Because this is a common function that is usually used, we need to make it as easy to use as possible. In this case, returning null means that the file couldn't be read.

ldhyen99 (approved):

PR #28, opened 2024-03-15
feat: message templates

huynguyen-hl:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📕 storybook


packages/untp-test-suite/src/templates/mapper.ts
```diff
@@ -0,0 +1,33 @@
...
+Handlebars.registerHelper('jsonArray', (array: object) => {
+  return JSON.stringify(array);
+});
+
+Handlebars.registerHelper('replace', (input: string, target: string, replacement: string) => {
+  const regex = new RegExp(target, 'g');
+  return input.replace(regex, replacement);
+});
+
+export async function templateMapper(templateName: string, testSuiteResult: TestSuiteResult) {
```
ldhyen99, line 20: Should we check `testSuiteResult` value if it exists or not?

huynguyen-hl: This project uses Typescript, so I don't think this function will execute if any of its parameters or nested fields are missing. Except if you call the built file (.js) of this file.

packages/untp-test-suite/src/templates/mapper.ts
```diff
@@ -0,0 +1,28 @@
...
+import { fileURLToPath } from 'url';
+import { dirname } from 'path';
+import { TestSuiteResult } from '../core/types/index.js';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+
+Handlebars.registerHelper('jsonArray', (array: object) => {
+  return JSON.stringify(array);
+});
```
namhoang1604, line 13: the name of the helper is not good. We should make general like `jsonStringify` because the input is an object. Furthermore, should we validate the input parameter?

packages/untp-test-suite/src/templates/utils/path.ts
```diff
@@ -0,0 +1,10 @@
+import { dirname } from 'path';
+import { fileURLToPath } from 'url';
+
+export const getCurrentFilePath = () => {
+  return fileURLToPath(import.meta.url);
+};
+
+export const getCurrentDirPath = (filePath: string) => {
+  return dirname(filePath);
```
namhoang1604, line 9: The code is created in the previous PR, should we utilize it and reuse it for the previous code?

huynguyen-hl: Yes, I agree. I've just updated the credentials utility functions to utilize this code, and then updated their unit tests as well.

namhoang1604 (approved):

ldhyen99 (approved):

PR #31, opened 2024-03-20
feat: integrate core test runner with mapper template

ldhyen99:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📕 storybook


packages/untp-test-suite/src/core/processTestSuite.ts
```diff
@@ -1,39 +1,93 @@
...
+  testSuiteResultAjv: IValidatedCredentials[],
+): Promise<ICredentialTestResult[]> => {
+  try {
+    const credentialsTemplate = [] as ICredentialTestResult[];
+    for (const value of testSuiteResultAjv) {
+      const templateName = getTemplateName(value);
+      const message = await templateMapper(templateName, value);
+
+      credentialsTemplate.push(JSON.parse(message));
+    }
```
namhoang1604, line 43: The loop handles asynchronous by `for` will run sequentially. Should we use `map` to run parallel?

namhoang1604 (changes requested):

packages/untp-test-suite/src/interfaces/utils/common.ts
```diff
@@ -9,4 +18,50 @@ export async function readJsonFile<T>(filePath: string): Promise<T | null> {
...
+  }
+
+  let finalStatus: FinalStatus = 'PASS';
+  if (finalMessage === TestSuiteMessage.Failed) {
+    finalStatus = 'FAILED';
+  }
+
+  if (finalMessage === TestSuiteMessage.Warning) {
+    finalStatus = 'WARN';
+  }
```
namhoang1604, line 61: should merge the condition expressions for `finalStatus` and `finalMessage`
And the condition expression for failed should be at the last condition, because if we have 3 credentials and the result of the test is [pass, failed, warning], then final status based on this code will be `'WARN'`, it should be failed

namhoang1604 (changes requested):

packages/untp-test-suite/src/interfaces/utils/common.ts
```diff
@@ -9,4 +18,50 @@ export async function readJsonFile<T>(filePath: string): Promise<T | null> {
...
+
+  if (credentials.some((credential) => credential.result === TemplateName.warn)) {
+    finalMessage = TestSuiteMessage.Warning;
+  }
+
+  let finalStatus: FinalStatus = 'PASS';
+  if (finalMessage === TestSuiteMessage.Failed) {
+    finalStatus = 'FAILED';
+  }
```
huynguyen-hl, line 57: Message template:
```
Testing Credential: productPassport v1.0
Path: path/to/credential.json
Result: PASS
---
Testing Credential: objectEvent v2.1
Path: path/to/objectEvent.json
Result: WARN - [Reason for warning (extended the core model)]
---
Testing Credential: aggregationEvent v2.1
Path: path/to/aggregationEvent.json
Result: FAIL - [Error details or reason for failure]
---
```

Final report:
```
|               UNTP Core Test Suite               |
|:------------------------------------------------:|
|               Runner version V0.01               |
| Credential Type        | Version | Status        |
|------------------------|---------|---------------|
| productPassport        | v1.2    | PASS          |
| conformityCredential   | v1.0    | PASS          |
| objectEvent            | v2.1    | PASS          |
| aggregationEvent       | v2.1    | PASS          |
| transformationEvent    | v2.1    | PASS          |
| transactionEvent       | v2.1    | PASS          |
|------------------------|---------|---------------|
|    **Your credentials are UNTP compliant 🎉**    |
```

I think we should follow the template we discussed before.
As you can see, both the message templates and the final report have three result/status types: `PASS`, `WARN`, `FAIL`. So, you can change `FAILED` to `FAIL` to match the template above.

packages/untp-test-suite/src/core/types/index.ts
```diff
@@ -25,4 +25,37 @@ export interface TestErrors {
...
+}
+
+export type FinalStatus = 'FAILED' | 'PASS' | 'WARN';
+
+export enum TemplateName {
+  failed = 'FAILED',
+  pass = 'PASS',
+  warn = `WARN`,
+  finalReport = 'finalReport',
+}
```
huynguyen-hl, line 55: We have created four template files: `error.hbs`, `pass.hbs`, `warning.hbs`, and `finalReport.hbs`. You can view the folder at `untp-test-suite/src/templates/templateMessages` for more detail. If you use this type enum here to load the exact template file, I think the enum values should be the same as the template file name.

You can refer to this:

```ts
export enum TemplateName {
  error = 'error',
  pass = 'pass',
  warning = `warning`,
  finalReport = 'finalReport',
}
```

packages/untp-test-suite/src/interfaces/utils/common.ts
```diff
@@ -9,4 +18,44 @@ export async function readJsonFile<T>(filePath: string): Promise<T | null> {
   } catch (error) {
     return null;
   }
-}
\ No newline at end of file
+}
+
+export const getTemplateName = (testSuiteResult: IValidatedCredentials): string => {
```
namhoang1604, line 23: I think `getTemplateName` and `generateFinalMessage` should be in the `templates` folder, these functions do not deal with the interface (CLI or Lib)

namhoang1604 (changes requested):

packages/untp-test-suite/src/interfaces/utils/common.ts
```diff
@@ -9,4 +18,44 @@ export async function readJsonFile<T>(filePath: string): Promise<T | null> {
...
+  if (testSuiteResult.errors === null) {
+    return TemplateName.pass;
+  }
+
+  if (Array.isArray(testSuiteResult.errors) && testSuiteResult.errors.length > 0) {
+    const getErrors = testSuiteResult.errors as ErrorObject[];
```
huynguyen-hl, line 29: I think variables that represent objects or entities should typically be named with nouns, while functions and methods should use verbs or action-oriented names.

packages/untp-test-suite/src/interfaces/utils/common.ts
```diff
@@ -1,4 +1,13 @@
 import { readFile } from 'fs/promises';
+import {
+  ICredentialTestResult,
+  IFinalReport,
+  TemplateName,
+  IValidatedCredentials,
+  TestSuiteMessage,
+  FinalStatus,
+} from '../../core/types';
```
huynguyen-hl, line 9: The `untp-test-suite` is a module package, so I think we need to include '.js' at the end of the file import lines.

packages/untp-test-suite/src/core/processTestSuite.ts
```diff
@@ -1,39 +1,84 @@
...
-  const credentialConfigs = await readJsonFile<ConfigCredentials>(credentialConfigsPath);
-  const validateCredentialResult = validateCredentialConfigs(credentialConfigsPath, credentialConfigs.credentials);
+  try {
+    // Read and validate credential configs
+    const credentialConfigs = await readJsonFile<ConfigCredentials>(credentialConfigsPath);
+    const validateCredentialResult = validateCredentialConfigs(credentialConfigsPath, credentialConfigs.credentials);
 
-  const configContentNoError = validateCredentialResult.filter((config) => config.errors === null);
+    // Filter out configs with no errors
+    const configContentNoError = validateCredentialResult.filter((config) => config.errors === null);
```
huynguyen-hl, line 55: I think the `validateCredentialConfigs` function, which is called at line 52, always returns a 2-dimensional array. This implies that the `config` parameter of the callback function inside this `filter` function never has an `errors` field with `null` data.

packages/untp-test-suite/src/core/processTestSuite.ts
```diff
@@ -1,39 +1,84 @@
...
```
huynguyen-hl, line 56: I think the `configContentNoError` variable here always contains an empty array. This is because the `config` parameter of the callback function inside this `filter` function never has an `errors` field with 'null' data.

You can review the details of the `validateCredentialConfigs` function (it's invoked at line 52), where you'll notice that the `errors` field is always an array.

packages/untp-test-suite/src/templates/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+      // If any ErrorObject has 'additionalProperty', return 'warning'
+      return TemplateName.warn;
+    }
+
+    // If no ErrorObject has 'additionalProperty', return 'FAIL'
+    return TemplateName.error;
+  } else {
+    return TemplateName.pass;
+  }
+};
```
huynguyen-hl, line 29: I think the first if statement should handle all cases to return the 'pass' template, and then the remaining code should only focus on returning the 'warn' or 'error' template.

This way, your code will be cleaner and easier to read. 

You can consider to this approach:

```ts
export const getTemplateName = (testSuiteResult: IValidatedCredentials): string => {
  if (!testSuiteResult.errors || !testSuiteResult.errors.length) {
    return TemplateName.pass;
  }

  const errors = testSuiteResult.errors as ErrorObject[];
  const isWarningType = errors.find((error) => error?.params?.additionalProperty);
  if (isWarningType) {
    // If any ErrorObject has 'additionalProperty', return 'warning'
    return TemplateName.warn;
  }

  // If no ErrorObject has 'additionalProperty', return 'FAIL'
  return TemplateName.error;
};
```

huynguyen-hl (dismissed):

namhoang1604 (approved):

namhoang1604 (approved):

huynguyen-hl (approved):

PR #32, opened 2024-03-20
feat: add 'testSuiteHandler' library test interface

huynguyen-hl:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📕 storybook

### DIFF
--- a/packages/untp-test-suite/src/interfaces/lib/testSuiteHandler.ts
+++ b/packages/untp-test-suite/src/interfaces/lib/testSuiteHandler.ts
@@ -1,0 +1,100 @@
+import { processTestSuite } from '../../core/processTestSuite.js';
+import { TestSuiteHandler } from './types.js';
+
+/**
+ * The purpose of the `testSuiteHandler` function is to read a JSON credential file specified by
+ * the `credentialPath` parameter. This file contains credential settings necessary to run
+ * the UNTP (UN Transparency Protocol) test suite.
+ * 
+ * The main functionality is to test your JSON data files to ensure they comply with the UNTP data model.
+ * @see Please see {@link https://uncefact.github.io/spec-untp/docs/specification/} for more information.
+
+ * The credential file typically includes a `credentials` field, which is an array containing JSON objects.
+ * Each object within this array describes the data file to be tested.
+ * 
+ * For example:
+ * {
+ *  "credentials": [
+ *    {
+ *      "type": "aggregationEvent",
+ *      "version": "v0.0.1",
+ *      "dataPath": "/data/aggregationEvent.json"
+ *    },
+ *    {
+ *      "type": "conformityEvent",
+ *      "version": "v0.0.1",
+ *      "dataPath": "/data/conformityEvent.json"
+ *    },
+ *    {
+ *      "type": "productPassport",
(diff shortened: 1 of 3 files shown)


packages/untp-test-suite/src/interfaces/lib/testSuiteHandler.ts
```diff
@@ -0,0 +1,310 @@
...
+ * // The 'credentialResults' variable will contain the test suite result.
+ */
+
+/**
+ * Handles testing of multiple credentials against their respective schemas and generates a final report.
+ * @param {string | ConfigCredentials} credentialOrCredentialPath - The path to the credential file or the actual ConfigCredentials object.
+ * @returns {Promise<TestSuiteResult>} A promise that resolves to the test results for all credentials and the final report.
+ */
+
+export const testMultiCredentialHandler: TestMultiCredentialHandler = async (credentialOrCredentialPath: string | ConfigCredentials): Promise<ITestMultiCredentialHandlerResult> => {
```
ldhyen99, line 211: I am curious about what's different between this function and processTestSuite() in the core.

namhoang1604: Agree. All the logic handles should be in the core layer. At the interface layer, it should wrap up and make up the result to be suitable for the interface.

namhoang1604 (changes requested):

ldhyen99 (approved):

packages/untp-test-suite/src/interfaces/lib/types.ts
```diff
@@ -0,0 +1,15 @@
+import { ErrorObject } from 'ajv';
```
ldhyen99, line 1: Please review this variable, it’s not used

ldhyen99 (approved):

packages/untp-test-suite/src/core/processTestSuite.ts
```diff
@@ -1,19 +1,41 @@
...
-export const processCheckDataBySchema = async (credentialConfig: ConfigContent): Promise<IValidatedCredentials> => {
+export const processCheckDataBySchema = async (
+  credentialConfig: IConfigContent,
+  data?: object,
+): Promise<IValidatedCredentials> => {
   const { type, version, dataPath } = credentialConfig;
-  const [schema, data] = await Promise.all([dynamicLoadingSchemaService(type, version), readJsonFile(dataPath)]);
+  let _data;
+
+  if (!data && !dataPath) {
```
ldhyen99, line 24: I think we should check data is {}. Because when I test with data is {} and without dataPath, the test case still shows a pass

namhoang1604: Your test schema does not require any properties, so it will pass when test the empty object

PR #43, opened 2024-04-29
docs: mock app configuration

ldhyen99:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [x] 📝 Documentation Update
- [ ] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [x] 📜 README.md
- [ ] 📕 storybook


ducpm511 (approved):

packages/mock-app/src/components/Router/Router.tsx
```diff
@@ -40,7 +40,7 @@ function Router() {
       <Route path='*' element={<Navigate to='/404' />} />,
       {/* Iterate through the appConfig to dynamically generate routes */}
       {appConfig.apps.map(renderRoutes)}
-      {appConfig.generateFeatures.map(renderRoutes)}
+      {appConfig.generalFeatures.map(renderRoutes)}
```
huynguyen-hl, line 43: Can you provide the latest version of `app-config.json` in the PR description? I noticed the `generateFeatures` field name has been updated to `generalFeatures`.

packages/services/src/storage.service.ts
```diff
@@ -1,58 +1,43 @@
...
-    await publicAPI.put(presignedUrl.toString(), file);
-
-    return `https://${bucket}.s3.ap-southeast-2.amazonaws.com/${file.name}`;
+    const getResultByPath = getStorageUrlByPath(result, params.resultPath);
+    return getResultByPath;
   } catch (error: any) {
-    throw new Error(error.message ?? 'Error uploading json');
+    console.error(error);
+    throw new Error(error.message);
   }
```
huynguyen-hl, line 26: I think this code block doesn't make sense. It can reduce the information of the error when the error is thrown to the caller of this function. So, I think this function doesn't need to use a try-catch block unless you want to customize the error message or add special logic in the catch block (e.g., debugging).

ldhyen99: Yes, after checking the function that calls `storageService` function, I think we'll refactor it later. Some events such as `Transformation Events`, it do not have try/catch. We cannot handle these errors without `try/catch` in `storageService`.

huynguyen-hl (approved):

PR #105, opened 2024-09-13
feat: integrate IDR with mock app

namhoang1604:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

This PR integrate the IDR with the mock app

https://github.com/gs-gs/fa-ag-trace/issues/561

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📕 storybook


packages/services/src/utils/helpers.ts
```diff
@@ -257,3 +257,8 @@ export const constructObject = (
 
   return data;
 };
+
+export const extractDomain = (url: string): string => {
```
ldhyen99, line 261: From my understanding, if we do not have try/catch when calling `new URL`, the default error is `Throws TypeError: Invalid URL`. Should we custom this error?

namhoang1604: No, we shouldn't

packages/services/src/utils/helpers.ts
```diff
@@ -301,6 +301,11 @@ export const constructObject = (
   return data;
 };
 
+export const extractDomain = (url: string): string => {
```
ldhyen99, line 304: Can you provide the unit test of this function?

namhoang1604: Good catch!

ldhyen99 (dismissed):

packages/services/src/linkResolver.service.ts
```diff
@@ -96,46 +101,51 @@ export interface GS1LinkResponse extends ILinkResponse {
...
 };
 
 export const constructLinkResolver = (
+  namespace: string,
   linkResolver: ILinkResolver,
   linkResponses: ILinkResponse[],
   qualifierPath: string,
 ) => {
-  const gs1LinkResolver: GS1LinkResolver = {
+  const LinkResolver: LinkResolver = {
```
huynguyen-hl, line 137: I think the `LinkResolver` variable should follow the camelCase naming convention.

packages/components/src/components/BarcodeGenerator/BarcodeGenerator.tsx
```diff
@@ -24,7 +24,38 @@ export const BarcodeGenerator = (props: IBarcodeProps) => {
...
+        const parsedValues = values.map((item: any) => {
+          const convertToGS1String = (obj: any) => {
+            if (typeof obj !== 'object' || obj === null) return '';
+
+            let result = '';
+
+            // Handle '01' (01) first if it exists
+            if ('01' in obj) {
+              let value = obj['01'];
+              if (!/^\d{8,14}$/.test(value)) throw new Error('GTIN must be 8-14 digits');
```
huynguyen-hl, line 36: The regex hardcoded in the `if` statement to check the `01` AI value here might be incorrect. I checked the GS1 codebase and noticed they use this regex `(\d{12,14}|\d{8})` to validate the `01` AI value. You may  double-check it to ensure it's correct.

namhoang1604: I checked and tested again, and both cases worked well but I changed to follow the GS1 regex

huynguyen-hl (approved):
LGTM

PR #127, opened 2024-10-22
fix: issue import EnvelopedVerifiableCredential

namhoang1604:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

This PR fixes the flow import EnvelopedVerifiableCredential

https://github.com/gs-gs/fa-ag-trace/issues/638

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [x] 📕 documentation

### DIFF
--- a/documentation/docs/mock-apps/components/render-check-list.md
+++ b/documentation/docs/mock-apps/components/render-check-list.md
@@ -17,7 +17,6 @@ The RenderCheckList component is responsible for rendering a list of items with 
   "name": "RenderCheckList",
   "type": "EntryData",
   "props": {
-    "requiredFieldPath": "/vc/credentialSubject/eventID",
     "checkBoxLabel": "Imported valid json:",
     "nestedComponents": [
       {
@@ -48,24 +47,118 @@ The RenderCheckList component is responsible for rendering a list of items with 
 
 ### Props
 
-| Property          | Required | Description                                                | Type                                       |
-| ----------------- | -------- | ---------------------------------------------------------- | ------------------------------------------ |
-| requiredFieldPath | Yes      | The path for the label of the checkbox item                | String                                     |
-| nestedComponents  | Yes      | An array of components to be rendered with the loaded data | [Component](/docs/mock-apps/components/)[] |
-| checkBoxLabel     | Yes      | The label for the checkbox list                            | String                                     |
-| style             | No       | The style for the component                                | Object                                     |
+| Property | Required | Description | Type |
+| ----------------- | -------- | ---------------------------------------------------------- | ------------------------------------------ |String |
+| nestedComponents | Yes | An array of components to be rendered with the loaded data | [Component](/docs/mock-apps/components/)[] |
+| checkBoxLabel | Yes | The label for the checkbox list | String |
+| style | No | The style for the component | Object |
 
 ## Response Data
 
-The component will return an object with data that is selected by the user from the list of checkboxes. The object will have the following structure:
+The data returned by the component will depend on the data type that is imported or scanned by the user. The data will be verified and decoded if it is an enveloped verifiable credential. There are 3 possible cases:
+
+1. If the imported or scanned data type is a `JSON`, the data will be returned without any transformation and have the key as the file name or URL of QRcode.
 
 ```json
 {
-  "0123456789": {
-    // Imported data
+  "data1.json": {
+    "foo": "bar" // Imported data
   },
-  "9876543210": {
-    // Imported data
+  "data2.json": {
+    "foo": "bar" // Imported data
+  }
+}
+```
+
+2. If the imported or scanned data type is a `VerifiableCredential`, the data will be returned as an object with the key as the file name or URL of QRcode. The imported or scanned VC will be verified based on the `vcOptions` provided in the [Import Button](./import-button) or [QR Code Scanner Dialog Button](./qr-code-scanner-dialog-button) component.
+
+```json
+{
+  "VC1.json": {
+    "@context": [], // Imported VC
+    "type": ["VerifiableCredential"]
+  },
+  "VC2.json": {
+    "@context": [], // Imported VC
+    "type": ["VerifiableCredential"]
+  }
+}
+```
+
+3. If the imported or scanned data type is a `VerifiableCredential` and the VC is an enveloped VC, the data will be returned as an object with the key as the file name or URL of QRcode. The imported or scanned VC will be verified and decoded based on the `vcOptions` provided in the [Import Button](./import-button) or [QR Code Scanner Dialog Button](./qr-code-scanner-dialog-button) component. The decoded VC will return the same level path as the original VC.
+   For example, if the original VC was imported as the whole object of the file, the decoded VC will be returned together with the original VC in an object like below:
+
+```json
+// imported VC file
+{
+  "@context": ["https://www.w3.org/ns/credentials/v2"],
+  "id": "data:application/vc-ld+jwt,jwt",
+  "type": "EnvelopedVerifiableCredential"
+}
+```
+
+```json
+// return data
+{
+  "VC1.json": {
+    "vc": {
+      "@context": ["https://www.w3.org/ns/credentials/v2"],
+      "id": "data:application/vc-ld+jwt,jwt",
+      "type": "EnvelopedVerifiableCredential"
+    },
+    "decodedEnvelopedVC": {
+      "@context": ["https://www.w3.org/ns/credentials/v2"],
+      "type": ["VerifiableCredential"],
+      "credentialSubject": {
+        "id": "did:example:123",
+        "name": "Alice"
+      }
+    }
+  }
+}
+```
+
+But when imported data has `credentialPath` in `vcOptions`, the decoded VC will be returned with the path specified in `credentialPath` together with the original VC in an object like below:
+
+```json
+// vcOptions
+{
+  "credentialPath": "/vc"
+}
+```
+
+```json
+// imported VC file
+{
+  "vc": {
+    "@context": ["https://www.w3.org/ns/credentials/v2"],
+    "id": "data:application/vc-ld+jwt,jwt",
+    "type": "EnvelopedVerifiableCredential"
+  },
+  "linkResolver": "https://example.com"
+}
+```
+
+```json
+// return data
+{
+  "VC1.json": {
+    "vc": {
+      "vc": {
+        "@context": ["https://www.w3.org/ns/credentials/v2"],
+        "id": "data:application/vc-ld+jwt,jwt",
+        "type": "EnvelopedVerifiableCredential"
+      },
+      "decodedEnvelopedVC": {
+        "@context": ["https://www.w3.org/ns/credentials/v2"],
+        "type": ["VerifiableCredential"],
+        "credentialSubject": {
+          "id": "did:example:123",
+          "name": "Alice"
+        }
+      }
+    },
+    "linkResolver": "https://example.com"
   }
 }
 ```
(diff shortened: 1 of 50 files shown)


packages/components/src/components/ImportButton/ImportButton.tsx
```diff
@@ -55,7 +66,16 @@ export const ImportButton = ({ label = 'Import', onChange}: IImportButtonProps)
       }
 
       const fileArray = Array.from(files);
-      const fileContents = await Promise.all(fileArray.map((file: File) => loadJsonFile(file)));
+      const fileContents = await Promise.all(
+        fileArray.map(async (file: File) => {
+          const content = await loadJsonFile(file);
+          if (type === ImportDataType.VerifiableCredential) {
+            const processedData = await processVerifiableCredentialData(content, vcOptions?.credentialPath);
+            return { [file.name]: processedData };
```
ldhyen99, line 74: If we return `[file.name]` field, I tried to integrate it with our event like `processDPP`, and the error as the image below. 
image

namhoang1604: The import and scanning functionality is currently only used for the processTransformationEvent. I forgot to remove the configuration in app-config, which caused some confusion here.

packages/services/src/vckit.service.ts
```diff
@@ -69,3 +76,41 @@ const constructCredentialObject = ({ context, type, issuer, credentialSubject, .
...
+ * @param vc
+ * @returns
+ */
+export const decodeEnvelopedVC = (vc: VerifiableCredential): UnsignedCredential | null => {
+  if (vc?.type === 'EnvelopedVerifiableCredential') {
+    const encodedCredential = vc?.id?.split(',')[1];
+    return decodeJwt(encodedCredential as string) as UnsignedCredential;
+  }
+  return null;
+};
```
huynguyen-hl, line 116: Should we use a try-catch block for this function?

packages/mock-app/src/pages/Verify.tsx
```diff
@@ -4,7 +4,7 @@ import { IVerifyResult, VerifiableCredential } from '@vckit/core-types';
 import { useLocation } from 'react-router-dom';
 import { computeEntryHash } from '@veramo/utils';
 import { Status } from '@mock-app/components';
-import { publicAPI, privateAPI } from '@mock-app/services';
+import { publicAPI, privateAPI, verifyVC } from '@mock-app/services';
```
huynguyen-hl, line 7: It seems the `verifyVC` function already wraps the API call to verify the VC, so I don’t think we need to import the `privateAPI` function here.

packages/mock-app/src/pages/Verify.tsx
```diff
@@ -95,17 +95,9 @@ const Verify = () => {
...
-        policies: {
-          credentialStatus: true,
-        },
-      };
-
       const verifyServiceUrl = appConfig.defaultVerificationServiceLink.href;
-      privateAPI.setBearerTokenAuthorizationHeaders(appConfig.defaultVerificationServiceLink.apiKey ?? '');
-      const verifiedCredentialResult = await privateAPI.post<IVerifyResult>(verifyServiceUrl, verifyCredentialParams);
+      const verifiedCredentialResult = await verifyVC(verifiableCredential, verifyServiceUrl);
+
```
huynguyen-hl, line 100: By default, the `verifyVC` function already retrieves the `appConfig.defaultVerificationServiceLink.href` from the `app-config.json`. So I don’t think we need to get the same `appConfig.defaultVerificationServiceLink.href` here.

packages/components/src/utils/importDataHelpers.ts
```diff
@@ -0,0 +1,30 @@
...
+        JSONPointer.set(data, credentialPath, vcData);
+      } else {
+        data = vcData;
+      }
+    }
+
+    return data;
+  } catch (error: any) {
+    throw new Error(error);
+  }
```
huynguyen-hl, line 29: I think the exception thrown here should provide a more descriptive error message to make debugging easier and to better understand the issue.

huynguyen-hl: If we keep it the same, sometimes this function will throw an error like:
`Uncaught Error: Error: Invalid Verifiable Credential!`

packages/components/src/components/RenderCheckList/RenderCheckList.tsx
```diff
@@ -30,18 +28,12 @@ export const RenderCheckList = ({ checkBoxLabel, requiredFieldPath, onChange, ne
...
-      const requiredFieldValue = getValueByPath(item, requiredFieldPath);
-      if (!requiredFieldValue) {
-        return acc;
-      }
-
-      return { ...acc, [requiredFieldValue]: item };
+      return { ...acc, ...item };
     }, {});
 
     return setCheckBoxes((prev) => ({ ...prev, ...validItems }));
```
huynguyen-hl, line 39: I'm not sure if this code works correctly when the `uploadedFiles` array contains multiple objects. Could you please double check it?

packages/services/src/vckit.service.ts
```diff
@@ -150,3 +153,41 @@ const constructCredentialObject = ({ context, type, issuer, credentialSubject, .
...
+): Promise<IVerifyResult> => {
+  const verifyCredentialParams = {
+    credential: verifiableCredential,
+    fetchRemoteContexts: true,
+    policies: {
+      credentialStatus: true,
+    },
+  };
+
+  privateAPI.setBearerTokenAuthorizationHeaders(appConfig.defaultVerificationServiceLink.apiKey ?? '');
```
ldhyen99, line 175: I have an error with this code:
image

packages/services/src/processDPP.service.ts
```diff
@@ -60,7 +62,7 @@ export const processDPP: IService = async (data: any, context: IDppContext): Pro
       deleteItemFromLocalStorage(context.localStorageParams);
     }
 
-    return { vc, linkResolver };
+    return { vc, decodedEnvelopedVC, linkResolver };
```
ldhyen99, line 65: Should we check the `linkResolver`? Because I can not verify it on the verify page
image

This is the error
image

packages/mock-app/src/pages/Verify.tsx
ashleythedeveloper: Hi @huynguyen-hl, this is an issue I have addressed in the latest commit to this branch, and this [PR](https://github.com/uncefact/tests-untp/pull/131) also sought to address.

packages/components/src/utils/importDataHelpers.ts
ashleythedeveloper: Hi @huynguyen-hl, how do you propose we address this? Are you thinking of having a generic error message instead of propagating the error message thrown or modifying the error messages thrown upstream?

packages/services/src/processDPP.service.ts
ashleythedeveloper: Hi @ldhyen99, I'm not sure I understand your comment. You are asking if we should validate the `linkResolver` object, but in your photos, it looks like the credential is not being registered with the storage service. Could you elaborate?

packages/components/src/components/ImportButton/ImportButton.tsx
ashleythedeveloper: @namhoang1604, but it can be used in other scenarios, right? For example, importing DCCs, DPPs, etc, into the system.

packages/mock-app/src/pages/Verify.tsx
ashleythedeveloper: Removed

packages/services/src/vckit.service.ts
ashleythedeveloper: Fixed in the latest commit.

packages/services/src/processDPP.service.ts
namhoang1604: I believe @ldhyen99 tested several times with this identity(the GTIN), so the responded URL was an old and removed stored DPP link. I think you need to clean your local services

ldhyen99: When I click on the `http://localhost:3005/gs1/gtin/05012345678900/10/2024001/TRF240001?linkType=gs1:certification` on localstorage, it redirects to verify page. In the `verify page` page, I see the error as in the image above, maybe this link was not stored in Storage Service.

packages/components/src/utils/importDataHelpers.ts
huynguyen-hl: This code can be improved for better error handling. Wrapping the error in a new Error object loses the original stack trace, which can make debugging more challenging. Instead, you can throw the original error or re-throw it to preserve its information:

```ts
} catch (error) {
    throw error; // re-throw the original error to preserve stack trace
}
```

Alternatively, if you need to add context to the error, you can include additional information without losing the original error’s details:

```ts
} catch (error: any) {
    throw new Error("Custom error message", { cause: error });
}
```

ashleythedeveloper: Okay. The issue, I believe, is that the stack trace is not preserved once the error is thrown.

In your first example, we might as well remove the try-catch block, right?

In the second example, yes, we preserve the stack trace but lose any well-crafted error messages in exchange for the custom message.

How do you think we could improve this, or is the custom message with the stack trace preserved sufficient?

huynguyen-hl: From my perspective, if we’re not adding more details to the error in the catch block, we might consider removing the try-catch block altogether.

namhoang1604: Thanks @huynguyen-hl and @ashleythedeveloper , I updated it

packages/components/src/__tests__/utils/importDataHelpers.test.tsx
```diff
@@ -0,0 +1,158 @@
...
+    const credentialPath = '/vc';
+
+    (verifyVC as jest.Mock).mockImplementation(() => ({
+      verified: false,
+    }));
+    (decodeEnvelopedVC as jest.Mock).mockReturnValue(null);
+
+    expect(async () => await processVerifiableCredentialData(data, vcContext, credentialPath)).rejects.toThrow(
+      'Invalid Verifiable Credential!',
+    );
```
huynguyen-hl, line 156: I noticed that you’ve changed the error message thrown when `processVerifiableCredentialData` falls into the catch block. I think this test case should be updated to expect the error message to match the new version.

huynguyen-hl (dismissed):

ldhyen99 (dismissed):

ldhyen99 (approved):

documentation/docs/mock-apps/components/render-check-list.md
```diff
@@ -48,24 +47,118 @@ The RenderCheckList component is responsible for rendering a list of items with
...
+        "type": "EnvelopedVerifiableCredential"
+      },
+      "decodedEnvelopedVC": {
+        "@context": ["https://www.w3.org/ns/credentials/v2"],
+        "type": ["VerifiableCredential"],
+        "credentialSubject": {
+          "id": "did:example:123",
+          "name": "Alice"
+        }
+      }
```
huynguyen-hl, line 159: Should we add a brief description of the `decodedEnvelopedVC` field here? I think this document would be better if we explained how it is generated.

huynguyen-hl (approved):

PR #157, opened 2024-11-12
feat: enable storage service encrypted endpoint

huynguyen-hl:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [x] 📝 Documentation Update
- [ ] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

This PR enables the `mock-app` to use the encrypted endpoint of the storage service. It also updates all process functions to utilize the `constructVerifyURL` function, so the `registerLinkResolver` function no longer handles URL construction logic, making the code clearer and easier to maintain. All related unit tests and documentation have been updated accordingly. Additionally, the `resultPath` setting in the storage configuration of `app-config.json` has been removed, as it is no longer needed.

https://github.com/gs-gs/fa-ag-trace/issues/694
https://github.com/gs-gs/fa-ag-trace/issues/695

- Using the unencrypted endpoint `(/v1/documents)`:

Issue DPP:

image
image

---

- Using the encrypted endpoint `(v1/credentials)`:

Issue DPP:

image
image

Failed to decrypt the encrypted VC:

image

Failed to compare the hash in the verify URL with the VC hash:

image

---

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [x] 📖 [mock-app docs site](https://uncefact.github.io/tests-untp/docs/mock-apps/)
- [ ] 📜 README.md
- [ ] 📕 storybook


namhoang1604 (dismissed):

documentation/docs/mock-apps/common/verify-link.md
ldhyen99, line 1: Should we add the example for uri & hash only in the payload?

packages/services/src/__tests__/aggregationEvent.test.ts
```diff
@@ -59,7 +64,8 @@ describe('processAggregationEvent', () => {
   it('should process aggregation event', async () => {
     (vckitService.issueVC as jest.Mock).mockImplementationOnce(() => aggregationVCMock);
     (vckitService.decodeEnvelopedVC as jest.Mock).mockReturnValue(aggregationVCMock);
-    (uploadData as jest.Mock).mockResolvedValueOnce(aggregationEventDLRMock);
+    (uploadData as jest.Mock).mockResolvedValueOnce({ uri: 'https://exampleStorage.com/vc.json', key: '123', hash: 'ABC123' });
+    (constructVerifyURL as jest.Mock).mockReturnValueOnce('http://localhost/event/1234');
```
ldhyen99, line 68: Hi, should we change the link return to follow the `verifyURL` return?
image

packages/services/src/__tests__/linkResolver.test.ts
```diff
@@ -30,6 +30,7 @@ describe('create link resolve service', () => {
 
     const mockValue = {
       eventLink: 'https://localhost/epcis-transformation-event/1234',
+      eventVerifyLink: 'https://verify.com/dev/verifyCredential?q=https://localhost/epcis-transformation-event/1234&key=123&hash=abcde123',
```
ldhyen99, line 33: Should we decode url?

documentation/docs/mock-apps/common/verify-link.md
huynguyen-hl: “I think the current example already covers this case, since the `Components` section explains that the `key` can be optional, so users understand that `key` can be omitted.

packages/services/src/__tests__/aggregationEvent.test.ts
huynguyen-hl: I understand you want the `constructVerifyURL` return the structure string contains the `/verify` string there, but I don't think we need to do that, since the `constructVerifyURL` function is a side effect function, we not test it in this test case, and the response from that function in this case that not valuable, you can see some lines below this line, where the data is respond by `constructVerifyURL` function is used by `registerLinkResolver` function, which already mocked. 

I understand that you want the `constructVerifyURL` function to return a structured string containing `/verify`, but I don’t think this is necessary. Since `constructVerifyURL` is a side effect function, we aren’t testing it directly in this case, and its response isn’t relevant here. Just a few lines below, you can see that the data returned by `constructVerifyURL` is used by the `registerLinkResolver` function, which is already mocked.

namhoang1604 (approved):

PR #160, opened 2024-11-20
fix: download vc button

huynguyen-hl:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [x] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

This PR adds the ability to display the `Download` button, enabling users to download the issued VC after clicking the issue (Submit) button, provided the `includeDownload` setting is enabled (set to true). It ensures users can download the complete content stored in local storage by the `mergeToLocalStorage` service in the previous step.

Additionally, this PR includes documentation for the `getValueFromLocalStorage` function, detailing how it works and how to use it effectively.

https://github.com/gs-gs/fa-ag-trace/issues/727

- Issue Transaction Event:

image

- Clicking the `Download` button:

image

- The downloaded content:

image

---

Docs site:

image
image

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [x] 📖 [Mock App docs site](https://uncefact.github.io/tests-untp/docs/mock-apps/)
- [ ] 📜 README.md
- [ ] 📕 storybook


ashleythedeveloper:
@huynguyen-hl, can we add some documentation around component types (void, submit, etc..) and what that means? For example, if I have a void component before download, it won't download due to the previous component being void. It might also be beneficial to add a little something to the CustomButton component as well.

namhoang1604 (approved):

ldhyen99 (approved):

huynguyen-hl:
> @huynguyen-hl, can we add some documentation around component types (void, submit, etc..) and what that means? For example, if I have a void component before download, it won't download due to the previous component being void. It might also be beneficial to add a little something to the CustomButton component as well.

@ashleythedeveloper, Yep, I agree with you. I’ll update all services to include the component types and provide a description of what they mean. Additionally, I’ll add more details to the CustomButton component documentation, explaining that if the last service in the services array is void, the button won’t be displayed because there is no data available for download.

namhoang1604 (dismissed):

documentation/docs/mock-apps/services/get-value-from-local-storage.md
```diff
@@ -0,0 +1,56 @@
+---
+sidebar_position: 28
```
ldhyen99, line 2: Is it a duplicate? If it's not, skip this comment

image

ldhyen99 (approved):

documentation/docs/mock-apps/services/get-value-from-local-storage.md
huynguyen-hl: Yep, it’s not duplicated in the `Services` tab. I think we only need to be concerned when using the same `sidebar_position` within the same parent tab.

PR #170, opened 2024-11-28
feat: add valid from valid until

ldhyen99:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

Add Support for validFrom/validUntil Fields in Root VC

Step to implement:
_ Check validUntil, if it is present in app-config.json, add it as a value field when issuing.
_ With validFrom:
  - If validUntil exists in parameters/vckitof app-config.json, the credential must be between validFrom and validUntil to be considered valid.
  - If validUntil does not exist, the credential is valid from validFrom onwards, which is the time issue the credential

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [x] 📖 [Mock App docs site](https://uncefact.github.io/tests-untp/docs/mock-apps/)
- [ ] 📜 README.md
- [ ] 📕 storybook


packages/services/src/vckit.service.ts
```diff
@@ -71,6 +71,19 @@ export const issueVC = async ({
     _validateVckitHeaders(headers);
   }
 
+  if (restOfVC?.validUntil) {
+    const validUntil = restOfVC.validUntil;
+    const validFrom = new Date().toISOString();
+
```
huynguyen-hl, line 77: I think we could allow users to input the validFrom field manually if they don’t want to use the current date and time. This would make the field more flexible.

namhoang1604 (dismissed):

huynguyen-hl (approved):

namhoang1604 (approved):

PR #204, opened 2025-01-29
fix(build): Copy templates to correct directory during yarn build.

absoludity:
## What type of PR is this? (check all applicable)

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

I just came across this while trying to fix something else:

Without this change, `yarn build` copies the templates to the wrong location after the first build, so any changes are not included:

When nothing exists, it works perfectly, creating the `templateMessages` directory with the files as expected:
```console
(venv) {fix-build-cp-of-templates} ~/dev/gosource/tests-untp/packages/untp-test-suite$ rm build/templates/templateMessages -rf
(venv) {fix-build-cp-of-templates} ~/dev/gosource/tests-untp/packages/untp-test-suite$ yarn build && ls -l build/templates/templateMessages
yarn run v1.22.22
$ tsc --build --clean && tsc && cp -r src/templates/templateMessages/ build/templates/templateMessages
Done in 1.41s.
total 16
-rw-rw-r-- 1 michael michael 134 Jan 29 12:48 credentialResult.hbs
-rw-rw-r-- 1 michael michael 337 Jan 29 12:48 errors.hbs
-rw-rw-r-- 1 michael michael  76 Jan 29 12:48 finalReport.hbs
-rw-rw-r-- 1 michael michael 247 Jan 29 12:48 warnings.hbs
```

but make a change to a template, and re-run `yarn build`, and you'll find your template changes aren't included, because they've been copied to `build/templates/templateMessages/templateMessages` . A confusing subtlety of `cp -r` :/ . 

```console
 yarn build && ls -l build/templates/templateMessages
yarn run v1.22.22
$ tsc --build --clean && tsc && cp -r src/templates/templateMessages/ build/templates/templateMessages
Done in 1.40s.
total 20
-rw-rw-r-- 1 michael michael  134 Jan 29 12:48 credentialResult.hbs
-rw-rw-r-- 1 michael michael  337 Jan 29 12:48 errors.hbs
-rw-rw-r-- 1 michael michael   76 Jan 29 12:48 finalReport.hbs
[...]

PR #208, opened 2025-01-29
fix(errors): error messages need to be escaped to ensure valid json.

absoludity:
If the error message includes a regex with, for example, backslashes (as is often the case with regexes), the untp test function fails to report on the results, instead erroring with invalid character when the error report is converted to a JSON object.

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

While working on a [proposal for more extensible JSON schema from jargon](https://github.com/absoludity/tests-untp/pull/1) I found that I couldn't run `yarn run untp test` with a schema that included certain regexes (specifically, the XMLSCHEMA11-2 regex for datetime), but instead, would get:

```console
 yarn run untp test
yarn run v1.22.22
$ node build/interfaces/cli/cli.js test
Run Untp test suites failed
SyntaxError: Bad escaped character in JSON at position 261
Done in 0.66s.
```

It took me way too long (partly my own fault) to realise it wasn't an issue in my JSON schema nor was it an out-dated ajv or something else. I eventually tracked it down to unescaped backslashes from the regex being included in error messages, which were then passed as JSON strings *without* being escaped.

First I thought I'd simply ensure the template used `jsonStringify` on the error message, but that didn't seem to work. Then I realised I needed to run `yarn build` to ensure the templates were copied over to the build dir (perhaps this could happen automatically if they're outdated, like Makefile functionality?). After another loss of time I realised my template change wasn't even being included when I ran `yarn build && yarn run untp test` because of #204 . Once I fixed that, I could see that my updated template was being copied, but I'd get an error:

```console
$ yarn build && yarn run untp test
yarn run v1.22.22
$ tsc --build --clean && tsc && cp -r src/templates/templateMessages/ build/templates/templateMessages
Done in 2.46s.
yarn run v1.22.22
$ node build/interfaces/cli/cli.js test
Run Untp test suites failed
Error: Failed to run mapper template. An error occurred in the Handlebars registerHelper 'jsonStringify' function. Please provide a valid JSON object.
Done in 0.65s.
```

So it turns out, unlike `JSON.stringify()`, the `jsonStringify` template helper doesn't apply to strings. I then realised this wasn't a standard handlebars filter but a locally defined helper (it's been a while), so I then updated it to support strings like `JSON.stringify()` does and saw that it worked IRL, but my test still failed... it turns out the test defines a replica of the errors.hbs template? Keen to understand why we do that rather than reading the actual template so we're testing that, rather than a duplicate which could be different over time?

Anyway, end of the long story is that with this change, I can test a schema with a regex with backslashes and get:

```console
 yarn run untp test
yarn run v1.22.22
$ node build/interfaces/cli/cli.js test
Testing Credential: digitalConformityCredential
Version: v0.5.0
Path: credentials/conformityCredential/DigitalConformityCredential_instance-v0.5.0.json
Result: FAIL
Error: 
/validFrom field called must match pattern '-?([1-9][0-9]{3,}|0[0-9]{3})-(0[1-9]|1[0-2])-(0[1-9]|[12][0-9]|3[01])T(([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9](\.[0-9]+)?|(24:00:00(\.0+)?))(Z|(\+|-)((0[0-9]|1[0-3]):[0-5][0-9]|14:00))'.
---
...
Done in 0.65s.
```

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📕 storybook

Nope

### DIFF
--- a/packages/untp-test-suite/__tests__/templates/mapper.test.ts
+++ b/packages/untp-test-suite/__tests__/templates/mapper.test.ts
@@ -21,7 +21,7 @@ describe('templateMapper', () => {
       {
         "fieldName": "{{instancePath}}",
         "errorType": "{{keyword}}",
-        "message": "{{instancePath}} field {{message}}.{{#if params.allowedValues}} Allowed values: {{{jsonStringify params.allowedValues}}}.{{/if}}"
+        "message": "{{instancePath}} field {{jsonStringify message}}.{{#if params.allowedValues}} Allowed values: {{{jsonStringify params.allowedValues}}}.{{/if}}"
       }
       {{#unless @last}},{{/unless}}
       {{/each}}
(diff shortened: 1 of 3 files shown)


packages/untp-test-suite/__tests__/templates/mapper.test.ts
```diff
@@ -21,7 +21,7 @@ describe('templateMapper', () => {
       {
         "fieldName": "{{instancePath}}",
         "errorType": "{{keyword}}",
-        "message": "{{instancePath}} field {{message}}.{{#if params.allowedValues}} Allowed values: {{{jsonStringify params.allowedValues}}}.{{/if}}"
+        "message": "{{instancePath}} field {{jsonStringify message}}.{{#if params.allowedValues}} Allowed values: {{{jsonStringify params.allowedValues}}}.{{/if}}"
```
absoludity, line 24: This is the change I was surprised I needed. Specifically, why do we have an exact copy of the errors.hbs template here in the test code, rather than just reading it in from the actual file so that we know we're testing the real deal? (I don't think the file access is significant enough to warrant an in-memory version, when compared with the danger of the file and the duplicate version here being out of sync). Let me know if I'm missing something.

ashleythedeveloper: @absoludity, I agree. There is no reason why this should be happening. Can you open another PR?

PR #212, opened 2025-02-06
feat: implement JSON-LD Document Expansion and Context Validation

huynguyen-hl:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

This PR aims to implement JSON-LD Document Expansion and Context Validation. The UNTP Playground now supports validating JSON-LD syntax, verifying remote context resolvability, checking for protected term conflicts, and ensuring that all properties are correctly expanded without undefined terms.

If credential validation fails due to context issues, a detailed view will be displayed, providing problem descriptions and guidance on how to resolve them.

https://github.com/gs-gs/fa-ag-trace/issues/908

Test that the properties aren’t defined in contexts:

image

Test for protected term conflicts:

Screenshot 2025-02-06 at 9 50 21 AM

Test invalid JSON-LD syntax:

Screenshot 2025-02-05 at 6 40 59 PM

Test remote context resolvability issues:

Screenshot 2025-02-06 at 9 50 06 AM

Unit tests cover all validation and checking functions for the context:

Screenshot 2025-02-06 at 2 26 14 PM

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📕 storybook


absoludity (approved):
Just looking at this because it's related to some things I've been looking at too.

The change looks excellent, thanks @huynguyen-hl . My only suggestion, or question, is whether it's worth including warnings about dropped fields. In particular, jsonld processing will drop any terms which do not expand to an iri, but as a warning, not an error. See my [explanation of where this affects the DCC](https://github.com/absoludity/spec-untp/blob/add-0.6.0-reasoning/docs/untp-data-model/changes-v0.6.0.md#deferred-some-terms-used-in-the-dcc-are-not-defined-in-our-imported-context), for example.

I think it would be worth highlighting these ignored fields, if present. Even though it's not an error for jsonld, I think it should be considered an error for us.

Let me know if that doesn't make sense.

Anyway, I"m approving this PR as is, but just noting that as a possible follow-up task.

BTW: It looks like we may be able to hook into the warning with an event handler - not sure:
https://github.com/digitalbazaar/jsonld.js/blob/71656073c2abe14b2538215ec8592cd9548b7a35/lib/expand.js#L488-L506

packages/untp-playground/src/components/TestResults.tsx
```diff
@@ -130,7 +131,7 @@ const TestStepItem = ({ step }: { step: TestStep }) => {
           <span>{step.name}</span>
         </div>
         {step.details &&
-          (step.id === 'schema' || step.id === 'extension-schema') &&
+          (step.id === 'schema' || step.id === 'context' || step.id === 'extension-schema') &&
```
ldhyen99, line 134: How do you think about using an Array with includes in this case?
For example: 
```

const allowedIds = ['schema', 'context', 'extension-schema'];
if (allowedIds.includes(step.id)) {
  // Your code here
}
```

huynguyen-hl: Your suggestion looks good. This PR slightly depends on [refactor PR](https://github.com/uncefact/tests-untp/pull/21), which contains similar logic at this line. I'll refactor the code following your suggestion after rebasing with the refactor PR.

ldhyen99 (dismissed):

huynguyen-hl:
> The change looks excellent, thanks @huynguyen-hl . My only suggestion, or question, is whether it's worth including warnings about dropped fields. In particular, jsonld processing will drop any terms which do not expand to an iri, but as a warning, not an error. See my [explanation of where this affects the DCC](https://github.com/absoludity/spec-untp/blob/add-0.6.0-reasoning/docs/untp-data-model/changes-v0.6.0.md#deferred-some-terms-used-in-the-dcc-are-not-defined-in-our-imported-context), for example.

> I think it would be worth highlighting these ignored fields, if present. Even though it's not an error for jsonld, I think it should be considered an error for us.

@absoludity, you’re right. It should be a warning, not an error. I’ll update this PR to highlight the ignored fields in the `View Details` tab. Thanks for your feedback!

absoludity:
Oh no, sorry!! I see now that your first image already highlights such properties (defined in the document but missing from the context). That's perfect. I'd missed that as I was looking for the same message I was seeing on the CLI. I think the way you have it already, as an error, is fine. Sorry for the noise.

huynguyen-hl:
Yep, no worries, it’s okay. You’ve given me a chance to take another look. I think the current error message is a bit confusing. I’ll update the error/warning message to make it clearer.

huynguyen-hl:
I've updated this PR based on the refactor PR merged by Ash. I also improved the error messages to make them clearer. This PR is now ready for review.

absoludity (approved):
Great, thanks @huynguyen-hl

PR #214, opened 2025-02-12
test: handle cleanup data cypress

ldhyen99:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

https://github.com/gs-gs/fa-ag-trace/issues/933

image

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📕 storybook


package.json
```diff
@@ -7,7 +7,7 @@
   ],
   "scripts": {
     "copy-config": "cp app-config.json packages/mock-app/src/constants/app-config.json && cp app-config.json packages/components/src/constants/app-config.json",
-    "copy-config-test": "cp cypress/fixtures/app-config.json packages/mock-app/src/constants/app-config.json && cp cypress/fixtures/app-config.json packages/components/src/constants/app-config.json",
+    "copy-config-test": "cp e2e/cypress/fixtures/app-config.json packages/mock-app/src/constants/app-config.json && cp e2e/cypress/fixtures/app-config.json packages/components/src/constants/app-config.json",
     "start": "yarn copy-config && cd packages/mock-app && yarn start",
     "start:e2e-test": "yarn copy-config-test && cd packages/mock-app && yarn start",
     "start:untp-playground": "cd packages/untp-playground && PORT=4001 yarn dev",
```
huynguyen-hl, line 13: Should we add scripts to trigger Cypress runs in package.json?

For example:

```json
"scripts": {
  // ...others
  "cypress:open": "cypress open",
  "cypress:run": "cypress run"
}
```

e2e/cypress.config.ts
```diff
@@ -34,16 +34,30 @@ export default defineConfig({
...
-              }
-              resolve(null);
-            });
-          });
+        resetData(file?: string) {
+
+          const targetDir = path.resolve(
+            process.cwd(),
+            `../minio_data/identity-resolver-service-object-store/data-test/idr-bucket-1/gs1${file ? `/${file}` : ''}`
+          );
```
huynguyen-hl, line 42: I don't think the `targetDir` variable will have the same value if we run Cypress from different working directories.

This means the code's behavior depends on where the user is when triggering Cypress, as `process.cwd()` returns the current working directory.

For example:
 If we run `process.cwd()` from the project's root directory, it might return: '/Users/user/tests-untp'. But if we run it from the cypress folder, it would return: '/Users/user/tests-untp/cypress'.

package.json
ldhyen99: Now, we can't because we've separated the end-to-end (e2e) tests from the root and created a separate package.json file for them.

image

e2e/cypress.config.ts
ldhyen99: Ah I see, your suggestions are good and I am thinking about it. 

Now our E2E test only runs on the E2E folder, although I trigger in the root, we still need to run `cd e2e` and start the test. How do you think about it? Do we have other cases?

e2e/cypress/e2e/issue_workflow_test/DCC/issueDCC.cy.ts
```diff
@@ -19,7 +18,7 @@ class DCCIssueFlow extends IssuePage {
 describe('Issue DCC end-to-end testing flow', () => {
   const dccTest = new DCCIssueFlow();
 
-  beforeEach(() => {
+  before(() => {
     dccTest.beforeAll();
   });
```
absoludity, line 23: Similar to my comment [here](https://github.com/uncefact/tests-untp/pull/214/files):

> It looks like we consistently clear the last credential before every single test? Could we instead run this in a `support/e2e.ts` or similar `before()` call, or do you think we might want some tests which don't clear the credential?
> 
> See https://docs.cypress.io/app/core-concepts/writing-and-organizing-tests#Support-file , especially the section with the
> 
> ```js
>   cy.log('I run before every test in every spec file!!!!!!')
> ```

In this case, we're calling `beforeAll()` in the `before` of each test case, but if you check the above link to the cypress docs, they have an example there where they use a single `beforeEach` from the **support** file, to show that it is then run before *any* test is run (which sounds like what we want here, rather than having to know we need to include this for each new test suite?)

ldhyen99: > Similar to my comment [here](https://github.com/uncefact/tests-untp/pull/214/files):
> 
> > It looks like we consistently clear the last credential before every single test? Could we instead run this in a `support/e2e.ts` or similar `before()` call, or do you think we might want some tests which don't clear the credential?
> > See https://docs.cypress.io/app/core-concepts/writing-and-organizing-tests#Support-file , especially the section with the
> > ```js
> >   cy.log('I run before every test in every spec file!!!!!!')
> > ```
> 
> In this case, we're calling `beforeAll()` in the `before` of each test case, but if you check the above link to the cypress docs, they have an example there where they use a single `beforeEach` from the **support** file, to show that it is then run before _any_ test is run (which sounds like what we want here, rather than having to know we need to include this for each new test suite?)

Thanks for the review. After reconsidering, I decided to revert before() to beforeEach() in the test of `issue_workflow_test` and move the data cleanup to a global after() in `support`. However, I noticed that some tests, like `untp-playground` or `actor_content_test`, do not require data cleanup. But this change does not affect those tests.

e2e/cypress/support/commands/common.ts
```diff
@@ -1,3 +1,4 @@
 after(() => {
-    cy.task('resetData');
-})
+  cy.task('resetData');
+  Cypress.env('lastCredential', undefined);
```
absoludity, line 3: Have you tested that this runs after each test? [From the name and the docs](https://docs.cypress.io/app/core-concepts/writing-and-organizing-tests#Hooks), `after()` runs once all tests are done, where as `afterEach` runs after each test block.

ldhyen99: Yes, I ran each test in Issue Flow, and the data was cleaned up.
image

Following the best practice of Cypress, we should change to before()
image

But I used before(), and I found that the data like gs1 in `minio` does not delete until the next test run. I am a little confused about it. Does it affect our testing or data?
image

image

absoludity: Yeah - whether it's `before` or `after` wasn't my question, but rather, whether it's `before` or `beforeEach`. Based on the link to the docs I sent above, I thought it'd have to be with the `Each`. Did you try `beforeEach()` as above?

e2e/cypress/e2e/issue_workflow_test/DCC/issueDCC.cy.ts
absoludity: Yes, but the question here is whether it should be just **once** in the `beforeEach()` of the support file (see link above), rather than needing to do it for each test case (4 times). That's what my question was about on the other PR. But anyway, it's a small detail to DRY up the code a bit. I'll +1 as is.

absoludity (approved):
See my inline comments

huynguyen-hl (approved):

e2e/cypress/support/commands/common.ts
ldhyen99: Yes, `beforeEach()` is good for cleaning up data, but in our test flow, the data needs to be preserved because we use the results from one test in the subsequent test.
image

absoludity: Ah - I see. I'd not realised that the test data is meant to persist between tests. Great, thanks :)

PR #216, opened 2025-02-14
test: add e2e test for transformation traceability event

ldhyen99:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

https://github.com/gs-gs/fa-ag-trace/issues/804

image

image

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📕 storybook


e2e/cypress/e2e/issue_workflow_test/Traceability/issueTransformationEvent.cy.ts
```diff
@@ -0,0 +1,42 @@
+import IssuePage from 'cypress/page/issuePage';
+
+class TransformationEventIssueFlow extends IssuePage {
+  testGenerateDPPWorkflow() {
+    this.generateWorkflow(
+      'Orchard Facility',
+      'Issue Traceability Event',
+      'DigitalTraceabilityEvent',
+      'DigitalTraceabilityEvent-v0.5.0.json',
```
absoludity, line 9: Should we be paramaterising the version here, or are we happy to be updating it. I see you've already paramaterised it below, why not here?

e2e/cypress/support/commands/issue-commands.ts
```diff
@@ -21,11 +21,16 @@ Cypress.Commands.add(
     const schemaURL = feature?.components[0].props?.schema?.url ?? '';
     const appService = feature?.services[0]?.parameters[0] ?? {};
 
-    cy.interceptAPI('GET', schemaURL, `get${schemaName}`);
+    if (schemaURL) {
+      cy.interceptAPI('GET', schemaURL, `get${schemaName}`);
```
absoludity, line 25: In what case is `schemaURL` not defined here? Won't this silently continue in that case? (is that what we want? I don't have enough context to know).

e2e/cypress/e2e/issue_workflow_test/Traceability/issueTransformationEvent.cy.ts
```diff
@@ -0,0 +1,42 @@
+import IssuePage from 'cypress/page/issuePage';
+
+class TransformationEventIssueFlow extends IssuePage {
+  testGenerateDPPWorkflow() {
```
absoludity, line 4: I'm confused why this sub-class is defining a `testGenerateDPPWorkflow` function, rather than a `DTE` specific workflow? I mean, if I look at:
- The cypress DPP workflow test it [defines and uses a `generateDPPWorkflow` function](https://github.com/uncefact/tests-untp/blob/181ddff4befbf51b3e40f1ce87e6d7deabc4eeb6/e2e/cypress/e2e/issue_workflow_test/DPP/issueDPP.cy.ts#L4-L12)
- The cypress DCC workflow test, it [defines and uses a `generateDCCWorkflow` function](https://github.com/uncefact/tests-untp/blob/181ddff4befbf51b3e40f1ce87e6d7deabc4eeb6/e2e/cypress/e2e/issue_workflow_test/DCC/issueDCC.cy.ts#L4-L12)
- The cypress DFR workflow test, it [defines and uses a `testGenerateDFRWorkflow` function](https://github.com/uncefact/tests-untp/blob/181ddff4befbf51b3e40f1ce87e6d7deabc4eeb6/e2e/cypress/e2e/issue_workflow_test/DFR/issueDFR.cy.ts#L4-L12)

That is, I **think** each should be defining a workflow specific to its use-case? It looks like that's what you're doing in the following 4 lines, so maybe it's just the name? (huh, similar for the `issueDIA.cy.ts` file, it seems to [define and use a `testGeneraeteDCCWorkflow` function](https://github.com/uncefact/tests-untp/blob/181ddff4befbf51b3e40f1ce87e6d7deabc4eeb6/e2e/cypress/e2e/issue_workflow_test/DIA/issueDIA.cy.ts#L4-L12), - but again, just looks like a copy-n-paste, the content looks specific to the DIA).

absoludity (commented):
A couple of small comments - just mis-named I think, but otherwise +1

e2e/cypress/support/commands/issue-commands.ts
ldhyen99: In `app-config.json`, we use [Json Form](https://jsonforms.io/) to render the UI from the schema.

image

The schema can be an object like this:
image

and It also can be a URL:
image

absoludity: Thanks for the context. It's still not clear to me, if `schema.url` was not defined because `schema` was an object, I don't see here in this context how it uses that schema object?

Anyway, I don't want to block you, so I'll +1 the review, but I am keen to understand what happens here if `components.schema` is an object here - from what I see here it looks like we only handle the case where `components.schema.url` is defined.

absoludity (approved):

e2e/cypress/support/commands/issue-commands.ts
absoludity: Oh - I have to resolve conversations for the PR to be able to be landed? (I'm normally keen to be able to unblock people while still continuing conversations for learning). But I'll resolve so it's not blocked.

e2e/cypress/e2e/issue_workflow_test/Traceability/issueTransformationEvent.cy.ts
absoludity: It's not clear to me how you resolved this either - that is, I see here that the version is included in the 'DigitalTraceabilityEvent-v0.5.0.json' arg, and in the new version you're instead passing the constant DIGITAL_TRACEABILITY_EVENT which is 'DigitalTraceabilityEvent', ie. without the version. I assume it's now also using the parameter below, passed in to `runUntpTest`.

Again, resolving this to unblock. Also I'm not sure if the 4000 line diff in `app-config.json` is intended for review, but I've not looked through it. Maybe next time if you could split mechanical changes into separate branch where it is clear it's not for review but a mechanical change (it looks like a lot of it is formatting, though I see there are some changes).

ashleythedeveloper (approved):

PR #227, opened 2025-02-28
feat: implement e2e tests into Github workflow

huynguyen-hl:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [x] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

This PR sets up Cypress E2E tests to run in both the `Build and Test` and `Release` workflows. The tests will be triggered whenever:

- A commit is pushed to the `main` branch (release).
- A PR is raised into the `next` branch.

Steps added to run E2E tests:

- Start E2E Docker Compose.
- Ensure IDR Service Readiness and Health (call the IDR health check API to ensure it’s ready to receive test requests).
- Run E2E tests.
- Stop Docker Compose.

In this PR, I’ve updated the `resetData` function to use the MinIO client library for deleting data in the `identity-resolver-service-object-store service` instead of the `fs` library for file system interactions. Additionally, I’ve renamed the function to `clearObjectStore` for better clarity. This removes the need to manually delete files or folders that are bind-mounted by the `identity-resolver-service-object-store` Docker Compose service, which would otherwise lead to a file system permission issue. Since the GitHub Actions environment runs under the `runner` user while the bind-mounted files and folders are owned by `root`, direct deletion is not possible.

Cypress E2E tests run successfully in a local environment because we use the same user (which has the highest privileges). However, if a different user were used locally, the same issue would occur.

### Testing E2E pipeline in Github personal account:

image

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📕 storybook


absoludity:
Hi @huynguyen-hl . This PR is in a Draft status (and has conflicts), and has some notes about things that aren't finished, so it's not clear if you're wanting a review yet?

> * Commented out `cy.task('resetData')` due to a file system permission conflict in the GitHub Actions environment. The `minio_data` folder is created by a Docker bind mount with root privileges, but the GitHub workflow runs under the runner user, which lacks permission to delete files inside `minio_data` (This is a temporary workaround until a better solution is implemented).

Can we not just `sudo` to delete the files from within the runner? See https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#administrative-privileges

huynguyen-hl:
Thanks @absoludity for the early feedback. I tried using `sudo`, but it still resulted in permission errors. I also used the `mxschmitt/action-tmate@v3` action to SSH into the GitHub Actions server for debugging and discovered that the root cause was the difference in folder ownership.

To address this, I took a different approach by using the MinIO client library to delete data from the `identity-resolver-service-object-store` service. The result is the same. It just uses a different method to handle the issue.

huynguyen-hl:
I noticed that some test cases in both `untp-playground/report-generation.cy.ts` and `untp-playground/vcdm-validation.cy.ts` are using unavailable files, such as:

- cypress/fixtures/credentials-e2e/valid-v2-enveloped-dpp.json
- cypress/fixtures/credentials-e2e/invalid-schema-v2-enveloped-dpp.json
- ...

This is why the E2E tests are failing.

ashleythedeveloper:
@

> I noticed that some test cases in both `untp-playground/report-generation.cy.ts` and `untp-playground/vcdm-validation.cy.ts` are using unavailable files, such as:
> 
> * cypress/fixtures/credentials-e2e/valid-v2-enveloped-dpp.json
> * cypress/fixtures/credentials-e2e/invalid-schema-v2-enveloped-dpp.json
> * ...
> 
> This is why the E2E tests are failing. I believe we'll need a separate ticket to address this issue.

@huynguyen-hl, this is the fix I was referring to in our last standup. I've raised PR [#229](https://github.com/uncefact/tests-untp/pull/229) to address this issue. I suggest to merge PR #229 first.

.github/workflows/build_test.yml
```diff
@@ -51,6 +51,52 @@ jobs:
             VC test suite, ./packages/vc-test-suite/coverage/coverage-summary.json
             UNTP Playground, ./packages/untp-playground/coverage/coverage-summary.json
 
+      - name: Start E2E docker compose
+        run: SEEDING=true docker compose -f docker-compose.e2e.yml up -d
+
+      - name: Ensure IDR Service Readiness and Health
```
ashleythedeveloper, line 57: Hi @huynguyen-hl, is there a reason why we aren't using the built-in [docker-compose health check](https://docs.docker.com/reference/compose-file/services/#healthcheck) functionality?

.github/workflows/release.yml
```diff
@@ -41,6 +41,52 @@ jobs:
       - name: Run tests
         run: yarn test
 
+      - name: Start E2E docker compose
+        run: SEEDING=true docker compose -f docker-compose.e2e.yml up -d
+
+      - name: Ensure IDR Service Readiness and Health
```
ashleythedeveloper, line 47: Same as above.

.github/workflows/build_test.yml
```diff
@@ -51,6 +51,52 @@ jobs:
             VC test suite, ./packages/vc-test-suite/coverage/coverage-summary.json
             UNTP Playground, ./packages/untp-playground/coverage/coverage-summary.json
 
+      - name: Start E2E docker compose
+        run: SEEDING=true docker compose -f docker-compose.e2e.yml up -d
```
ashleythedeveloper, line 55: Will GitHub cache the docker image between each run? If so, we need to build the image so the new functionality is present in the containers.

.github/workflows/release.yml
```diff
@@ -41,6 +41,52 @@ jobs:
       - name: Run tests
         run: yarn test
 
+      - name: Start E2E docker compose
+        run: SEEDING=true docker compose -f docker-compose.e2e.yml up -d
```
ashleythedeveloper, line 45: Same as above.

e2e/cypress.config.ts
```diff
@@ -3,9 +3,21 @@ import { defineConfig } from 'cypress';
 import fs from 'fs';
 import path from 'path';
 import util from 'util';
+import { Client, ClientOptions } from 'minio';
 
 const execPromise = util.promisify(exec);
 export default defineConfig({
+  env: {
+    idrBucketName: process.env.IDR_BUCKET_NAME || 'idr-bucket-1',
```
ashleythedeveloper, line 11: I'm not sure why we are diverging from the [existing environment variable names](https://github.com/uncefact/tests-untp/blob/587f0c5a0e44f17ddd64b793cfef40c0e3d0b52c/docker-compose.e2e.yml#L50) for the IDR services.

.github/workflows/build_test.yml
huynguyen-hl: As you can see, when we run docker compose up, it pulls all the Docker images defined in the compose file and starts the services. The terminal/console will return the cursor to us, allowing us to run another command, regardless of whether health checks are set up for each service. However, in the background, docker compose up continues running to initialize all services.

At this point, the next step—running the E2E tests—will be triggered. But how do the E2E tests determine whether the services have started successfully and passed their internal health checks as defined in the Docker Compose file? The docker compose up command does not wait until all services are fully running and healthy before returning control to us. So how can we ensure that all services are ready before running the E2E tests?

.github/workflows/release.yml
huynguyen-hl: Same as above.

.github/workflows/build_test.yml
huynguyen-hl: Do you mean that GitHub Actions has the capability to cache all Docker images defined in the docker-compose file? If so, I'm not sure about that. I believe it can cache a specific standalone Docker image, but I'm not sure if it applies to all images in a docker-compose setup.

If your idea isn’' what I thought, please share more details. Thank you!

e2e/cypress.config.ts
huynguyen-hl: Yep, you're right. I'll update these names to match the ENVs of IDR service definitions in docker-compose.e2e.yml.

.github/workflows/build_test.yml
ashleythedeveloper: Ah, I've misunderstood how the health check within docker compose works. The health check is non-blocking.

.github/workflows/build_test.yml
ashleythedeveloper: Yes, I mistyped. I meant "images".

e2e/cypress.config.ts
```diff
@@ -3,9 +3,21 @@ import { defineConfig } from 'cypress';
 import fs from 'fs';
 import path from 'path';
 import util from 'util';
+import { Client, ClientOptions } from 'minio';
 
 const execPromise = util.promisify(exec);
 export default defineConfig({
+  env: {
+    idrBucketName: process.env.OBJECT_STORAGE_BUCKET_NAME || 'idr-bucket-1',
+    idrGS1Prefix: process.env.OBJECT_STORAGE_GS1_PREFIX || 'gs1',
```
ldhyen99, line 12: Hi, what if I have the `nlis` Prefix in here? So I'll add another value in `.env` file, doesn't it?

e2e/cypress.config.ts
```diff
@@ -16,7 +28,7 @@ export default defineConfig({
       runMode: 2, // Retries in headless mode
       openMode: 0, // No retries in interactive mode
     },
-    defaultCommandTimeout: 4000,
+    defaultCommandTimeout: 10000,
```
ldhyen99, line 31: Why do we set `10000` here? Do you calculate something or is it just an assumption?

package.json
```diff
@@ -45,6 +45,7 @@
     "@types/uuid": "^9.0.7",
     "cross-env": "^7.0.3",
     "cypress": "^13.16.1",
+    "minio": "^8.0.4",
```
ldhyen99, line 48: Should we install `minio` in `e2e` package instead of in the root? I saw that we only call it in `cypress`

.github/workflows/build_test.yml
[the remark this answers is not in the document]
huynguyen-hl: Yep, it has been removed.

e2e/cypress.config.ts
huynguyen-hl: It has been removed.

e2e/cypress.config.ts
huynguyen-hl: It has been returned to its original value

package.json
huynguyen-hl: Updated.

ldhyen99 (approved):

ashleythedeveloper (approved):
LGTM

Issue #224, opened 2025-02-25
Add More Detailed Error Guidance for Non-UNTP Credentials


ashleythedeveloper:
When uploading credentials to the UNTP Playground that are not properly formatted for UNTP, users receive insufficient information to guide them on required next steps. The error messages lack specific details about formatting issues and remediation steps.

## Current Issues
1. When uploading an array of credentials (incorrect format), the system does not clearly indicate that each credential should be uploaded as a separate file.

2. After separating credentials into individual files, the system identifies that it cannot determine the credential type, but fails to explain specifically what needs to be added (a permitted UNTP credential type to the Type array).

3. When adding a valid UNTP credential type (e.g., "DigitalProductPassport"), users receive a "failed to load schema" message without explanation of why it failed (missing UNTP context IRIs).

## Example
[credential.json](https://github.com/user-attachments/files/18973813/credential.json)

<img width="1782" alt="Image" src="https://github.com/user-attachments/assets/55326507-1303-4f39-b598-bb2bd3c7443b" />
<img width="1782" alt="Image" src="https://github.com/user-attachments/assets/7bcd598f-5dfb-461d-ae77-9422a724b904" />
<img width="1782" alt="Image" src="https://github.com/user-attachments/assets/99240d9b-d176-4be8-95ce-e98d6ac6e112" />

## Proposed Solution
Implement detailed, step-by-step error messages that:
- Identify the specific formatting issue encountered
- Explain what the correct format should be
- Provide clear guidance on how to remediate the error
- Where possible, offer examples of correct formatting

The goal is to create a clear roadmap that guides users toward UNTP compliance, particularly when they are attempting to adapt credentials from other traceability pilots.

ldhyen99:
This is the UI and the message I propose to address these issues.

> When uploading an array of credentials (incorrect format), the system does not clearly indicate that each credential should be uploaded as a separate file.

<img width="1440" alt="Image" src="https://github.com/user-attachments/assets/ff78824d-295c-4556-a6bc-c54d6d77daf0" />

> After separating credentials into individual files, the system identifies that it cannot determine the credential type, but fails to explain specifically what needs to be added (a permitted UNTP credential type to the Type array).

My message proposes the following template:

Clarity: Make the error immediately understandable (what went wrong).

Guidance: Provide actionable steps (how to fix it).

Readability: Format the message so it’s easy to scan, even if detailed.

I am considering the display time of this notice, should we add the time, or should the user click on the close button?
[...]

PR #230, opened 2025-03-03
feat: improve UNTP playground error messages

ldhyen99:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

https://github.com/uncefact/tests-untp/issues/224

image
image

image
image

image
image

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📕 storybook

### DIFF
--- a/e2e/cypress/e2e/untp-playground/upload-invalid-credential.cy.ts
+++ b/e2e/cypress/e2e/untp-playground/upload-invalid-credential.cy.ts
@@ -1,0 +1,87 @@
+describe('Display Error Messages', () => {
+  beforeEach(() => {
+    cy.visit('http://localhost:4000');
+  });
+
+  it('should show detail popup if upload file is not an object', () => {
+    cy.uploadCredential([]);
+    cy.contains('Fix validation error').click();
+    cy.contains('Expected type: object').should('be.visible');
+    cy.contains('Receive value:').should('be.visible');
+    cy.contains('[]').should('be.visible');
+    cy.contains('Issue:').should('be.visible');
+  });
+
+  it('should view detail of error when `UNTP Schema Validation` failed', () => {
+    const invalidContextCredential = {
+      '@context': ['https://www.w3.org/ns/credentials/v2'],
+      type: ['VerifiableCredential', 'DigitalProductPassport'],
+      issuer: {
+        id: 'did:example:123',
+        name: 'dev',
+      },
+      credentialSubject: {
+        name: 'John Doe',
+        id: 'did:example:123',
+        type: ['Product'],
+      },
+    };
+
+    cy.uploadCredential(invalidContextCredential);
+    cy.expandGroup();
+    cy.checkValidationStatus('UNTP Schema Validation', 'failure');
+
+    cy.openErrorDetailsByStepName('UNTP Schema Validation');
+    cy.contains('Fix validation error').click();
+    cy.contains('Issue:').should('be.visible');
+  });
+
+  it('should open error details when clicking on View Detail Upload', () => {
+    cy.uploadCredential([]);
+    cy.contains('Close').click();
+    cy.contains('View Upload Detail').click();
+    cy.contains('Fix validation error').click();
+    cy.contains('Expected type: object').should('be.visible');
+    cy.contains('Receive value:').should('be.visible');
+    cy.contains('[]').should('be.visible');
+    cy.contains('Issue:').should('be.visible');
+  });
+
+  it('should display error message when upload multiple files', () => {
+    cy.uploadCredential([
+      [],
+      {
+        '@context': ['https://www.w3.org/ns/credentials/v2'],
+        type: ['VerifiableCredential'],
+        issuer: {
+          id: 'did:example:123',
+          name: 'dev',
+        },
+        credentialSubject: {
+          name: 'John Doe',
+          id: 'did:example:123',
+          type: ['Product'],
+        },
+      },
+      {
+        '@context': ['https://www.w3.org/ns/credentials/v2', 'https://test.uncefact.org/vocabulary/untp/dpp/0.5.0/'],
+        type: ['VerifiableCredential', 'DigitalProductPassport'],
+        issuer: {
+          id: 'did:example:123',
+          name: 'dev',
+        },
+        credentialSubject: {
+          name: 'John Doe',
+          id: 'did:example:123',
+          type: ['Product'],
+        },
+      },
+    ]);
+
+    cy.contains('Fix validation error').click();
+    cy.contains('Close').click();
+    cy.expandGroup();
+    cy.checkValidationStatus('JSON-LD Document Expansion and Context Validation', 'success');
+    cy.contains('View Upload Detail').click();
+  });
+});
--- a/e2e/cypress/e2e/untp-playground/upload-invalid-credential.cy.ts
+++ b/e2e/cypress/e2e/untp-playground/upload-invalid-credential.cy.ts
@@ -47,7 +47,7 @@ describe('Display Error Messages', () => {
     cy.contains('Issue:').should('be.visible');
   });
 
-  it('should display error message when upload multiple files', () => {
+  it('handles multiple file uploads displaying errors or success for each', () => {
     cy.fixture('credentials-e2e/valid-v2-enveloped-dpp.json').then((credential) => {
       cy.uploadCredential([
         [],
--- a/e2e/cypress/e2e/untp-playground/upload-invalid-credential.cy.ts
+++ b/e2e/cypress/e2e/untp-playground/upload-invalid-credential.cy.ts
@@ -48,35 +48,25 @@ describe('Display Error Messages', () => {
   });
 
   it('should display error message when upload multiple files', () => {
-    cy.uploadCredential([
-      [],
-      {
-        '@context': ['https://www.w3.org/ns/credentials/v2'],
-        type: ['VerifiableCredential'],
-        issuer: {
-          id: 'did:example:123',
-          name: 'dev',
+    cy.fixture('credentials-e2e/valid-v2-enveloped-dpp.json').then((credential) => {
+      cy.uploadCredential([
+        [],
+        {
+          '@context': ['https://www.w3.org/ns/credentials/v2'],
+          type: ['VerifiableCredential'],
+          issuer: {
+            id: 'did:example:123',
+            name: 'dev',
+          },
+          credentialSubject: {
+            name: 'John Doe',
+            id: 'did:example:123',
+            type: ['Product'],
+          },
         },
-        credentialSubject: {
-          name: 'John Doe',
-          id: 'did:example:123',
-          type: ['Product'],
-        },
-      },
-      {
-        '@context': ['https://www.w3.org/ns/credentials/v2', 'https://test.uncefact.org/vocabulary/untp/dpp/0.5.0/'],
-        type: ['VerifiableCredential', 'DigitalProductPassport'],
-        issuer: {
-          id: 'did:example:123',
-          name: 'dev',
-        },
-        credentialSubject: {
-          name: 'John Doe',
-          id: 'did:example:123',
-          type: ['Product'],
-        },
-      },
-    ]);
+        credential,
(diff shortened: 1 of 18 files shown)


packages/untp-playground/src/lib/utils.ts
```diff
@@ -58,3 +58,17 @@ export const downloadJson = (data: Record<string, any>, filename: string) => {
...
+ * This function uses `Object.prototype.toString` to get the internal
+ * [[Class]] property of the value, which is a more reliable way to
+ * determine the type of an object than using `typeof` or `instanceof`.
+ *
+ * The result is a string such as "String", "Number", "Array", "Object", etc.
+ *
+ * @param value - The value whose type is to be determined.
+ * @returns The type of the value as a string.
+ */
+export const typeOf = (value: any) => Object.prototype.toString.call(value).slice(8, -1);
```
huynguyen-hl, line 74: As I understand it, this function has the same functionality as the built-in `typeof` function.

Are you saying that using strings like "String", "Number", "Array", and "Object" makes the output easier to read compared to the default typeof results?

huynguyen-hl (changes requested):
image

When I upload an empty file, the `ValidationDetailsSheet` dialog opens automatically with the `View Details` button is still visible behind it (see the picture).

huynguyen-hl (dismissed):

e2e/cypress/support/commands/untp-playground.ts
```diff
@@ -2,7 +2,22 @@
...
+      const fileObject = {
+        contents: Buffer.from(JSON.stringify(cred)),
+        fileName: 'credential.json', // default file name; customize if needed
+        mimeType: 'application/json',
+      };
+      credentials.push(fileObject);
+    });
+
+    cy.get('[data-testid="credential-upload-input"]').selectFile(credentials, { force: true });
+  } else if (typeof credential === 'object') {
```
absoludity, line 20: This appears to be updating the `uploadCredential` helper so that if `credential` is an array (of length > 0) then the `credentials` variable will be an array of separate file references (separate file uploads)? Where as, based on the issue #224, the credential we're trying to handle is a **single file** which has a json array of credentials.

So, for example, I can currently upload multiple files because the playground explicitly allows and handles that fine (it seems?):

Screenshot from 2025-03-18 13-12-59

which then checks them all - great:

Screenshot from 2025-03-18 13-13-17

The above code appears to simulate that situation which at first made me think the change which I haven't gotten to yet, will stop this current working behaviour that allows multiple credential files, rather than the one on #224 which is a **single file** upload that is a json array with multiple credentials.

But checking below it looks like the change does what is intended by the issue (verifies that any one single file upload can't be an array of credential objects). So I think this test isn't testing what you think it's testing, which might be related to my first comment above in `upload-invalid-credential.cy.ts`. I'll update my comment there.

packages/untp-playground/src/app/page.tsx
```diff
@@ -20,13 +24,25 @@ export default function Home() {
...
   }>({});
+  const [fileCount, setFileCount] = useState(0);
+  const { dispatchError, errors, setIsDetailsOpen } = useError();
+
+  const shouldDisplayUploadDetailBtn =
+    errors &&
+    errors.length > 0 &&
+    ((fileCount === 1 && Object.keys(testResults).length === 0) ||
+      (fileCount > 1 && Object.keys(testResults).length !== 0) ||
+      (fileCount > 1 && Object.keys(testResults).length === 0));
```
absoludity, line 35: Aren't these two lines the same as: `fileCount > 1` ? (I mean, either the length is not equal to 0 or it's equal to zero)

e2e/cypress/e2e/untp-playground/upload-invalid-credential.cy.ts
```diff
@@ -0,0 +1,87 @@
...
+          type: ['Product'],
+        },
+      },
+    ]);
+
+    cy.contains('Fix validation error').click();
+    cy.contains('Close').click();
+    cy.expandGroup();
+    cy.checkValidationStatus('JSON-LD Document Expansion and Context Validation', 'success');
+    cy.contains('View Upload Detail').click();
```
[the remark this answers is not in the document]
absoludity: Actually, based on the next file (see my comment there), I'm pretty sure this test shows 'Fix validation error' only because you've provided *invalid* credentials, not because you've provided multiple credentials (in separate files). Can you check this by ensuring that all 3 credentials above are actually valid, so we can see if you get the error you expect?

e2e/cypress/support/commands/untp-playground.ts
ldhyen99: Yes, I discovered that we can upload multiple credentials. I have written the code and tests to handle and validate these credentials. This approach is different from handling a single credential where the value is an array.

e2e/cypress/e2e/untp-playground/upload-invalid-credential.cy.ts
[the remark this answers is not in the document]
ldhyen99: In this scenario, I upload three credentials, two of which are invalid, to trigger the error sheet. Using the "Fix validation error", I want to ensure that the sheet displays two error messages. After verifying the errors, I close the sheet and check the status of a valid credential.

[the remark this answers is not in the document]
absoludity: Ah - so this test is actually demonstrating that multiple credential files can be uploaded at once? I mistook it for a new test demonstrating the actual issue. Maybe updating the to something like:

```
it("handles multiple file uploads displaying errors or success for each", () => {
```

so that it's clear what is being tested? Fine either way.

e2e/cypress/support/commands/untp-playground.ts
absoludity: Yep, sorry, I mistook the test for a demonstration of the issue. I'll continue the review then :) Thanks!

packages/untp-playground/src/components/CredentialUploader.tsx
```diff
@@ -1,15 +1,31 @@
...
+
   const onDrop = useCallback(
     (acceptedFiles: File[]) => {
+      resetErrors();
       const validExtensions = ['.json', '.jwt', '.txt'];
+      if (acceptedFiles.length > 1) {
+        setFileCount(acceptedFiles.length);
+      } else {
+        setFileCount(1);
+      }
```
absoludity, line 28: What is logically different between the above 5 lines and:
```suggestion
      setFileCount(acceptedFiles.length);
```

packages/untp-playground/src/components/ErrorDialog.tsx
```diff
@@ -48,7 +54,28 @@ const groupErrors = (errors: any[]) => {
   };
 };
 
-export const ErrorDialog = ({ errors = [], className = '' }) => {
+const getTipMessage = (mainError: ValidationError) => {
```
absoludity, line 57: Thanks for extracting that out to a function - much neater below.

packages/untp-playground/src/lib/utils.ts
```diff
@@ -58,3 +58,47 @@ export const downloadJson = (data: Record<string, any>, filename: string) => {
...
+      params: {
+        type: 'object',
+        receivedValue: normalizedCredential,
+        solution: 'Instead of [credential1, credential2], upload credential1.json and credential2.json.',
+      },
+      message: 'Credentials must be uploaded as separate files, not as an array.',
+    };
+  }
+
+  if (!normalizedCredential || typeof normalizedCredential !== 'object') {
```
absoludity, line 90: So you're still using the built-in typeof operator, rather than your new function `typeOf`, which isn't used at all, so let's delete it?

ldhyen99: Yes, I deleted the function `typeOf`

packages/untp-playground/src/lib/utils.ts
```diff
@@ -58,3 +58,46 @@ export const downloadJson = (data: Record<string, any>, filename: string) => {
...
+ *
+ * This function uses `Object.prototype.toString` to get the internal
+ * [[Class]] property of the value, which is a more reliable way to
+ * determine the type of an object than using `typeof` or `instanceof`.
+ *
+ * The result is a string such as "String", "Number", "Array", "Object", etc.
+ *
+ * @param value - The value whose type is to be determined.
+ * @returns The type of the value as a string.
+ */
```
absoludity, line 73: Looks like you deleted the function but left the docstring for it? That might be confusing for someone reading this later... can we remove it too?

absoludity (commented):
Just needs the docstring for the now removed typeOf function to also be removed, otherwise +1

absoludity (dismissed):

packages/untp-playground/src/components/TestResults.tsx
```diff
@@ -164,35 +165,29 @@ const TestStepItem = ({ step }: { step: TestStep }) => {
...
-                  </SheetHeader>
-                  <div className='mt-4 overflow-y-auto max-h-[calc(100vh-8rem)]'>
-                    {step.details.errors && step.details.errors.length > 0 ? (
-                      <ErrorDialog errors={step.details.errors} className='w-full max-w-none' />
-                    ) : (
-                      <div className='text-yellow-600'>
-                        <p>⚠️ Additional properties found in credential</p>
-                      </div>
-                    )}
+        {step.details && isAllowedTestCase && shouldShowDetails && (
```
ashleythedeveloper, line 168: Is there a reason why we shouldn't use the `ValidationDetailsSheet` component you've created?

packages/untp-playground/src/components/TestResults.tsx
```diff
@@ -505,6 +500,13 @@ export function TestResults({ credentials, testResults, setTestResults }: TestRe
                             keyword: 'schema',
                             message: 'Failed to fetch schema',
                             instancePath: '',
+                            params: {
+                              missingValue: 'The schema could not be loaded due to missing UNTP context IRIs.',
+                              solution:
+                                "Ensure the credential includes the required UNTP context IRIs in the '@context' field.",
```
ashleythedeveloper, line 506: Just FYI, this error can also occur if the playground fails to fetch the schema for the extension.

packages/untp-playground/src/components/TestResults.tsx
```diff
@@ -520,7 +522,9 @@ export function TestResults({ credentials, testResults, setTestResults }: TestRe
 
           if (!validateContextResult.valid) {
             allChecksPass = false;
-            toast.error(validateContextResult.error!.message);
+            toast.error(
```
ashleythedeveloper, line 525: Do we really need this toast?

packages/untp-playground/constants.ts
```diff
@@ -58,3 +58,7 @@ export enum TestCaseStepId {
   EXTENSION_SCHEMA_VALIDATION = 'extension-schema-validation',
   CONTEXT_VALIDATION = 'context',
 }
+
+export const allowedContextValue = {
+  '@context': ['https://www.w3.org/ns/credentials/v2', 'https://test.uncefact.org/vocabulary/untp/{type}/{version}/'],
```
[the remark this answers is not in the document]
ldhyen99: Should we make these updates in this PR or create a separate PR? I noticed some code still references the 'test.uncefact.org' domain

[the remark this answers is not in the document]
ashleythedeveloper: We can handle this in a subsequent PR.

packages/untp-playground/src/components/TestResults.tsx
ldhyen99: According to me, besides reporting errors, it also informs the user that the process has been completed.

ashleythedeveloper (approved):

Issue #247, opened 2025-03-12
Testing extensions should not stop if extension not recognised


absoludity:
**Describe the bug**

I was just trying to test a DLP release with the following file:

[aatp-dlp-instance-with-bovineanimal-0.4.2-beta1.json](https://github.com/user-attachments/files/19218789/aatp-dlp-instance-with-bovineanimal-0.4.2-beta1.json)

I wasn't sure if it would be recognised but I did expect to see it checked as a valid DigitalProductPassport.

Instead, the test doesn't even start and a toast pops briefly saying "Unknown credential type".

NOTE: It could be bad data - I've not checked it thoroughly yet or debugged locally with a local playground.

Now the type of the credential is:

```
  "type": [
    "DigitalLivestockPassport",
    "DigitalProductPassport",
    "VerifiableCredential"
  ],
```

and I've just taken a quick look at the playground code and [found the `detectCredentialType` function](https://github.com/uncefact/tests-untp/blob/7d34d05dc89b0b9037affdffffe5e2173045bb69/packages/untp-playground/src/lib/credentialService.ts#L23-L34). This appears to return the **first** matching type from the credential that it recognises, which is also not great, though probably left over from when a DLP wasn't a DPP as well. Now that it is, we should recognise that a credential is both a DLP and a DPP, for example. Anyway, that's unrelated I think: the issue here that I've not had the time to check is why it's not recognised. Sorry if it is an issue with the data.

**To Reproduce**
1. Go to https://test.uncefact.org/test-untp-playground
2. Click on the add new credential box
3. Select the file attached above (after having downloaded it)
4. See the error.

**Expected behavior**

Unless there is a problem with my data, I expect the credential to be recognised as a DLP and a DPP.

PR #248, opened 2025-03-16
feat: add dlp v0.4.2-beta1 to extension config

ashleythedeveloper:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description
This PR adds the AATP DLP v0.4.2-beta1 release to the extension config

Related to #247, but doesn't resolve the underlying issue.

Screenshot 2025-03-17 at 9 55 42 am

## Added tests?

- [ ] 👍 yes
- [x] 🙋 no, because I need help

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📕 storybook

### DIFF
--- a/packages/untp-playground/src/lib/schemaValidation.ts
+++ b/packages/untp-playground/src/lib/schemaValidation.ts
@@ -42,6 +42,11 @@ export const EXTENSION_VERSIONS: Record<string, ExtensionConfig> = {
         schema: 'https://aatp.foodagility.com/schema/aatp-dlp-schema-0.4.1-beta1.json',
         core: { type: 'DigitalProductPassport', version: '0.6.0-beta7' },
       },
+      {
+        version: '0.4.2-beta1',
+        schema: 'https://aatp.foodagility.com/schema/aatp-dlp-schema-0.4.2-beta1.json',
+        core: { type: 'DigitalProductPassport', version: '0.6.0-beta9' },
+      },
     ],
   },
 };


absoludity (approved):
+1, though it's not clear to me yet why we even need to manually add the versions of extensions hard-coded here? Ah - perhaps it's mainly a map so we know where to find the extension's schema? (though why does it need to specify the exact core version - wouldn't that be linked from the data anyway?)

ashleythedeveloper:
> +1, though it's not clear to me yet why we even need to manually add the versions of extensions hard-coded here? Ah - perhaps it's mainly a map so we know where to find the extension's schema?

@absoludity, yes, the intent is to provide a mapping between the extension and the core credential type, as well as the corresponding versions.

> (though why does it need to specify the exact core version - wouldn't that be linked from the data anyway?)

I believe you are referring to using the context link present in the DLP to determine the core UNTP data model and version of which the extension is extending. Do I have that right?

If so, I agree this is a much more dynamic approach, and I believe the reason for not doing this is due to the absence of the core UNTP context link in the previous versions of the DLP.

We have since realised, by your guidance, that this is technically incorrect and have since rectified the issue. This would allow us to implement the above.

PR #266, opened 2025-05-13
fix: untp playground build

### LINKED ISSUE #265: UNTP Playground Container not Building
**Describe the bug**
When building the containers using Docker Compose (standard/e2e), the process fails when it attempts to build the UNTP Playground container. This has been observed locally and within the build/test pipeline.

The issue can be observed in [this pipeline run](https://github.com/uncefact/tests-untp/actions/runs/14896095529/job/41838817040?pr=231).

**To Reproduce**
1. Navigate to the repository
2. Run `SEEDING=true docker-compose up -d --build`

**Expected behavior**
The containers are built and running on their designated port.

ashleythedeveloper:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

This PR resolves a build failure in the UNTP Playground when provisioning the service locally or in the pipeline using Docker/Docker Compose (see issue #265).

The issue arose because the Dockerfile referenced the Yarn lock file in the `untp-playground` package directory, but the repository structure places the Yarn lock file in the root directory, with dependencies listed in the `package.json` file within the packages directory.

I identified three possible solutions:

**Option 1**  
Copy the Yarn lock file into the `untp-playground` package within the workflow file. This simple solution fixes the pipeline issue but leaves local builds broken.

**Option 2**  
Use the Docker Compose [additional_contexts](https://docs.docker.com/reference/compose-file/build/#additional_contexts) attribute to include the root directory and reference it in the Playground’s Dockerfile to copy the Yarn lock file. While straightforward, this requires the latest Docker Compose version, which would complicate documentation.

**Option 3 (chosen)**  
Change the build context. The Yarn lock file lies outside the build context set in the Docker Compose file (`packages/untp-playground`). I updated the build context to the root directory and modified the Dockerfile to copy the lock file from the root and the `untp-playground` package.

I've also refined the copy command to only copy existing files and removed the wildcard suffix from `yarn.lock`. This ensures the pipeline fails if the lock file is missing.

### Additional Fix
I've also included a fix for the Cypress environment (I know, naughty me). The end-to-end (e2e) test step was failing due to a Node version update in GitHub Action workers, which has apparently enabled experimental features by default. These features caused issues when transpiling the Cypress config written in TypeScript (see [pipeline run](https://github.com/uncefact/tests-untp/actions/runs/14987679978/job/42104577783) and [issue](https://github.com/cypress-io/github-action/issues/1408)). The fix sets two experimental feature flags to `false` in the workflow:

```yaml
env:
  NODE_OPTIONS: "--no-experimental-require-module --no-experimental-detect-module"
```
Which resolves the issue (see [pipeline run](https://github.com/uncefact/tests-untp/actions/runs/14988287837/job/42106273807)).

The remaining e2e test failures in the pipeline will be resolved once the reference implementation’s data models are updated to v0.6.0.

fixes #265

Screenshot 2025-05-13 at 1 41 29 pm
Screenshot 2025-05-13 at 12 16 53 pm
Screenshot 2025-05-13 at 12 17 21 pm

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📕 storybook

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -62,6 +62,8 @@ jobs:
 
       - name: Run E2E tests
         run: yarn test:run-cypress
+        env:
+          NODE_OPTIONS: "--no-experimental-require-module --no-experimental-detect-module" # See issue https://github.com/cypress-io/github-action/issues/1408
 
       - name: Stop docker compose
         run: docker compose -f docker-compose.e2e.yml down
--- a/docker-compose.e2e.yml
+++ b/docker-compose.e2e.yml
@@ -1,7 +1,8 @@
 services:
   untp-playground:
     build:
-      context: packages/untp-playground
+      context: .
+      dockerfile: packages/untp-playground/Dockerfile
     ports:
       - '4000:3000'
     volumes:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,7 +14,8 @@ services:
 
   untp-playground:
     build:
-      context: packages/untp-playground
+      context: .
+      dockerfile: packages/untp-playground/Dockerfile
     ports:
       - '4000:3000'
     volumes:
--- a/packages/untp-playground/Dockerfile
+++ b/packages/untp-playground/Dockerfile
@@ -12,7 +12,7 @@ ENV NEXT_PUBLIC_BASE_PATH=/
 ENV NEXT_PUBLIC_ASSET_PREFIX=/
 
 # Install dependencies based on the preferred package manager
-COPY package.json yarn.lock* package-lock.json* pnpm-lock.yaml* .npmrc* ./
+COPY package.json ../../yarn.lock* ./
 RUN yarn install
 
 
--- a/packages/untp-playground/Dockerfile
+++ b/packages/untp-playground/Dockerfile
@@ -12,7 +12,7 @@ ENV NEXT_PUBLIC_BASE_PATH=/
 ENV NEXT_PUBLIC_ASSET_PREFIX=/
 
 # Install dependencies based on the preferred package manager
-COPY package.json ../../yarn.lock* ./
+COPY packages/untp-playground/package.json yarn.lock ./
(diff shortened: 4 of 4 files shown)


.github/workflows/build_test.yml
```diff
@@ -62,6 +62,8 @@ jobs:
 
       - name: Run E2E tests
         run: yarn test:run-cypress
+        env:
+          NODE_OPTIONS: "--no-experimental-require-module --no-experimental-detect-module" # See issue https://github.com/cypress-io/github-action/issues/1408
```
absoludity, line 66: Thanks for including the ref - much easier to delete things like this again if it's clear where to check.

absoludity (dismissed):
Excellent - thanks for helping me learn by documenting the PR well.

Question: given that the yarn.lock file lives in the root directory, it'll be modified by `yarn install`ing other directories within this root directory right? I guess that's not an issue because if you're working on the playground, it'll have been the last one touching the yarn.lock file?

absoludity:
I've not checked the current build failure.

ashleythedeveloper:
> Question: given that the yarn.lock file lives in the root directory, it'll be modified by `yarn install`ing other directories within this root directory right? I guess that's not an issue because if you're working on the playground, it'll have been the last one touching the yarn.lock file?

Good catch! 

With Yarn workspaces, `yarn install` shouldn't alter the lock file unless packages are modified (e.g., added or removed). 

However, I missed the `--immutable` flag in the playground's Dockerfile, which would have resulted in a non-deterministic environment. 

I've updated the Dockerfile to use `yarn install --immutable` for consistent installations.

Issue #306, opened 2025-07-07
untp-test-suite uses redundant `credentials` file which is awkward for a UX as an API


absoludity:
The `untp test` command of the `untp-test-suite` package requires a --config arg for a `credentials` file which is really just specifying the files you want tested, much like a directory of files would, or individual file paths would.

The structure of the credentials file is:

```json
{
  "credentials": [
    {
      "type": "digitalConformityCredential",
      "version": "v0.5.1",
      "dataPath": "credentials/conformityCredential/DigitalConformityCredential_instance-v0.5.0.json",
      "url": ""
    },
    ...
  ]
}
```

AFAICT, the only useful bit of information is really the `dataPath`. Both the type and version can be obtained from the data (assuming it can be parsed, and it's an error if it can't anyway). The `url` property is also redundant as a `path` property can be both a local file or a URI, if needed.

So I'm proposing, similar to the args I've enabled in the https://github.com/absoludity/untp-graph-validation-cli POC, that we enable testing individual files as straight arguments, like:

`untp test path/to/file1.json path/to/file2.json`

as well as a directory of files:

`untp test --dir path/to/many/files/`

This API would then be much simple to re-use in the UX (playground), where we will want the user to be able to both upload individual files as well as upload an archive of multiple files, without the playground needing to construct a `credentials.json` file to re-use the test suite.

absoludity:
So I've done a few PRs towards this goal, more as a chance to learn the tests-untp repo, but they update the internal API to no longer depend on a credential config (though adding a task to the memory bank for a higher-level config to map credential types to schemas, only required for extensions of UNTP).

I'm at the point of refactoring the actual validation internals themselves, re-organising into tier 1 (and adding jsonld validation etc.) and tier 2 (and updating the dynamic loader to be more useful - loading a different schema for the same credential based on the type argument (ie. so you can get the VCDM schema, or the DPP schema, or the DLP schema when extensions are supported). But I'm going to hold off moving forward as I've a couple of questions there. The main one is whether we'd be best just copying the complete playground lib directory to start and using that (I think Ashley has given that more attention and it already does many of the things we'd need), or the untp-graph-validation-api lib. Either way, I think it would be better than refactoring and adding to what's there. Plus we could at the same time be updating the playground to import and use the version from untp-tests as we go.

So I'll wait until we get to discuss these options before moving forward here.

PR #308, opened 2025-07-14
feat: Remove dependence on `type` property of credential config

absoludity:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

This PR is step 1 of a few, towards removing the need for a credentials config file (#306) so that we have an API which is just a list of paths (to credential files) which can be re-used easily in other contexts (such as the playground).

Specifically, this PR removes the dependence on the `type` property of the config file. I've still allowed it to be present (it's optional) so this change shouldn't break existing setups, but it is ignored, with the `type` of the actual credential referred to by the path being used instead.

If you agree with the direction, I'll follow up with a PR removing the need for version and url and remove the config file entirely, as per #306 .

#306 

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [x] 📜 README.md
- [ ] 📕 storybook

Not that I'm aware of.


absoludity:
Closing in favour of #314

Issue #352, opened 2025-12-16
Define service interfaces in packages/services


ashleythedeveloper:
The credential issuance workflow involves three discrete operations: issuing (signing) a credential, storing the credential, and publishing the credential link. Each of these should be abstracted behind a well-defined interface so that implementations can be swapped without changing consuming code.

These interfaces act as contracts that define the expected behaviour without coupling to a specific implementation. Both the interfaces and their implementations should reside within the packages/services package.

### Acceptance Criteria:
- [ ] Define IVerifiableCredentialService interface with a method to issue/sign a credential payload and return a signed verifiable credential.
- [ ] Define IStorageService interface with a method to store a credential and return the storage URI and hash.
- [ ] Define IIdentityResolverService interface with a method to register a credential link against an identifier/scheme.
- [ ] All interfaces should have clearly typed inputs and outputs.
- [ ] Export interfaces and associated types from the services package.

PR #360, opened 2025-12-16
feat: define service interfaces

bla-ce:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

This PR defines service interfaces in `package/services` so that implementations can be swapped without changing consuming code.

ref #352 

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📕 storybook

### DIFF
--- a/packages/services/src/interfaces/identityResolverService.ts
+++ b/packages/services/src/interfaces/identityResolverService.ts
@@ -1,33 +1,138 @@
 /**
- * RFC 9264 compliant Linkset type - https://datatracker.ietf.org/doc/rfc9264/
- * A linkset is a collection of links described using the link context object
+ * UNTP Identity Resolver Types
+ *
+ * These types model the UN Transparency Protocol's Identity Resolver specification,
+ * which extends established standards for link resolution.
+ *
+ * Base standards:
+ * - ISO 18975: Link resolution framework
+ * - RFC 9264: Linkset format (returned by resolver queries)
+ * - GS1 Digital Link: Inspiration for identifier resolution patterns
+ *
+ * UNTP extensions:
+ * - Decentralised access control (encryption, access roles)
+ * - UNTP-specific link relation types (untp:dpp, untp:dcc, untp:dte)
+ *
+ * @see https://untp.unece.org/docs/specification/IdentityResolver
+ * @see https://untp.unece.org/docs/specification/DecentralisedAccessControl
+ * @see https://www.iso.org/standard/85540.html
+ * @see https://datatracker.ietf.org/doc/rfc9264/
+ * @see https://www.gs1.org/standards/gs1-digital-link
+ *
+ * @example
+ * // Publish a public DPP and a restricted DCC for an ABN
+ * const links: Link[] = [
+ *   {
+ *     href: 'https://storage.example.com/dpp-123.json',
+ *     rel: 'untp:dpp',
+ *     type: 'application/ld+json',
+ *     title: 'Digital Product Passport',
+ *     hreflang: ['en'],
+ *     default: true
+ *   },
+ *   {
+ *     href: 'https://storage.example.com/dcc-456.json',
+ *     rel: 'untp:dcc',
+ *     type: 'application/ld+json',
+ *     title: 'Conformity Certificate',
+ *     hreflang: ['en'],
+ *     method: 'POST',
+ *     encryptionMethod: 'AES-256-GCM',
+ *     accessRole: [AccessRole.Auditor]
+ *   }
+ * ];
+ *
+ * const registration = await identityResolver.publishLinks('abn', '51824753556', links);
+ * // => { resolverUri: 'https://resolver.example.com/abn/51824753556', ... }
  */
-export type Linkset = LinkContextObject[];
 
-export interface LinkContextObject {
-  anchor?: string;
-  [linkRelationType: string]: LinkTarget[] | string | undefined;
+/**
+ * UNTP access roles for variant-based disclosure.
+ *
+ * @see https://untp.unece.org/docs/specification/VariantBasedDisclosure
+ */
+export enum AccessRole {
+  /** Public access - any party holding a decryption secret may access */
+  Anonymous = 'untp:accessRole#Anonymous',
+  /** Business partners and purchasers */
+  Customer = 'untp:accessRole#Customer',
+  /** Competent authorities */
+  Regulator = 'untp:accessRole#Regulator',
+  /** Accredited recycling facilities */
+  Recycler = 'untp:accessRole#Recycler',
+  /** Authorized auditors */
+  Auditor = 'untp:accessRole#Auditor',
 }
 
-export interface LinkTarget {
+/**
+ * RFC 9264 link target attributes.
+ *
+ * @see https://datatracker.ietf.org/doc/rfc9264/
+ */
+export type RFC9264Link = {
+  /** URL of the linked resource */
   href: string;
-  type?: string;
-  hreflang?: string | string[];
-  title?: string;
-  'title*'?: Record<string, string>;
-  media?: string;
-}
+  /** Link relation type (e.g., "untp:dpp", "untp:dcc", "untp:dte") */
+  rel: string;
+  /** IANA Media Type of the target resource (e.g., "application/ld+json") */
+  type: string;
+  /** Human-readable title for the link */
+  title: string;
+  /** Language(s) of the target resource (e.g., ["en"], ["en", "de"]) */
+  hreflang?: string[];
+};
 
-export interface LinkRegistration {
-  resolverUri: string;      // Canonical URI where this identifier can be resolved
+/**
+ * UNTP-specific link extensions for decentralised access control.
+ *
+ * @see https://untp.unece.org/docs/specification/DecentralisedAccessControl
+ */
+export type UNTPLinkExtensions = {
+  /** HTTP method required to access the resource (default: GET) */
+  method?: 'GET' | 'POST';
+  /** Encryption method if the resource is encrypted (e.g., "AES-256-GCM") */
+  encryptionMethod?: string;
+  /** Access roles required to access the resource */
+  accessRole?: AccessRole[];
+  /** Whether this is the default link for its relation type */
+  default?: boolean;
+};
+
+/**
+ * A link to be published to the identity resolver.
+ * Combines RFC 9264 link attributes with UNTP-specific extensions.
+ */
+export type Link = RFC9264Link & UNTPLinkExtensions;
+
+/**
+ * Result of publishing links to an identity resolver.
+ */
+export type LinkRegistration = {
+  /** Canonical URI where this identifier can be resolved */
+  resolverUri: string;
+  /** The identifier scheme (e.g., "abn", "nzbn", "lei") */
   identifierScheme: string;
+  /** The identifier value within the scheme */
   identifier: string;
-}
+};
 
+/**
+ * Service responsible for publishing links to an identity resolver.
+ *
+ * Identity resolvers allow identifiers (e.g., ABN, NZBN, LEI) to be resolved
+ * to their associated resources. When queried, the resolver returns an
+ * RFC 9264 compliant linkset.
+ *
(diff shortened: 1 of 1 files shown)


packages/services/src/interfaces/storageService.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+export interface IStorageService {
+  /**
+   * Stores a signed verifiable credential using the provided storage configuration
+   */
+  store(
+    config: StorageConfig, 
+    crendential: SignedVerifiableCredential
+  ): Promise<StorageResponse>
+}
```
absoludity: This method would require some kind of token to be able to store the credential, wouldn't it? I don't see where that is (can't see it in the `StorageConfig`?

packages/services/src/interfaces/verifiableCredentialService.ts
```diff
@@ -0,0 +1,126 @@
...
+ */
+export interface IVerifiableCredentialService {
+/**
+   * Issues a verifiable credential
+   */
+  issue(
+    config: IssueConfig, 
+    payload: CredentialPayload
+  ): Promise<SignedVerifiableCredential>
+}
```
absoludity: Similar question about what authz is used to issue the credentials?

bla-ce:
@absoludity @ashleythedeveloper When writing tests for the IVerifiableCredentialService interface, I realised it might be better not to have 'config' as an argument. It feels like it should not depend on the configuration. Let me know if you agree and I'll quickly refactor.

We would use only what we need from the configuration and pass it as arguments. For example:

```typescript
  sign(
    config: IssueConfig, 
    payload: CredentialPayload
  ): Promise<SignedVerifiableCredential>
```

to:

```typescript
  sign(
    url: string,
    headers: Record<string, string>,
    payload: CredentialPayload
  ): Promise<SignedVerifiableCredential>

```

packages/services/src/interfaces/storageService.ts
absoludity: Sorry, what I meant was that this actual request function would need to include the token - ah, or you mean that the implementation will access the token directly (from the env or app config) - ok, I hadn't realised it has access to the implementation has access to the app config without the data being passed in. Great.

absoludity:
> @absoludity @ashleythedeveloper When writing tests for the IVerifiableCredentialService interface, I realised it might be better not to have 'config' as an argument. It feels like it should not depend on the configuration. Let me know if you agree and I'll quickly refactor.
> 
> We would use only what we need from the configuration and pass it as arguments. For example:
> 
> ```ts
>   sign(
>     config: IssueConfig, 
>     payload: CredentialPayload
>   ): Promise<SignedVerifiableCredential>
> ```
> 
> to:
> 
> ```ts
>   sign(
>     url: string,
>     headers: Record<string, string>,
>     payload: CredentialPayload
>   ): Promise<SignedVerifiableCredential>
> ```

Yep, I agree - the function should (IMO) have the minimal info as args that it requires to execute (which, btw, is why I expected the token to be passed in - but I'm OK if you want the implementations to pull those out, or if you're using a client that automatically injects them or whatever).

absoludity:
@ashleythedeveloper I've reviewed this but I don't know enough about RI to confidently +1, but I will +1 so @bla-ce is unblocked to land it. If you've time, please do take a look as you've more context here (but no problem if you've too much on).

@bla-ce I assume you're seeing the test failure? (I've not checked if it's related or not, but let's fix it either way before landing if possible)

absoludity (dismissed):
I'd already approved this PR before I went on break? I see you'd dismissed my approval (I assume because you added changes that you thought invalidated the review?) FWIW, I'll generally add a +1 with a caveat like I did here so that you aren't blocked from landing.

Anyway, +1 to removing the config. I'm unsure about the `headers?` arg that's been added to `store` and `sign` - iff that's being used to provide the authz for the calls. Personally I think a caller of those functions shouldn't need to know how to structure the header authz, and instead should just be able to pass the authz token and have the function add the required header. But that's a small point, nothing to block on.

+1 (again :P)

packages/services/src/interfaces/storageService.ts
```diff
@@ -0,0 +1,42 @@
...
+} from "@/types";
+
+import {
+  SignedVerifiableCredential
+} from "./verifiableCredentialService";
+
+/**
+ * Result of storing a verifiable credential in external storage
+ */
+export type StorageResponse = {
```
ashleythedeveloper: I'm not sure about extending `StorageResponse` with `Extensible`.

TL;DR: I think we should keep `StorageResponse` strict, and only extend via explicit, typed mechanisms so we preserve a stable contract.

I understand the intent: it gives us flexibility to add new fields over time, or to let a storage provider return implementation specific metadata.

However, I think this works against the purpose of having a service agnostic abstraction in the first place. The interface is meant to define a stable, well defined contract for downstream consumers. Implementations should be free to do whatever they need internally, but the output should conform to a consistent response shape that consumers can rely on.

Allowing arbitrary extension fields makes the contract open ended and encourages consumers to start depending on provider specific properties. Over time, that pushes logic into “if implementation X then use field Y”, which couples downstream code to a particular backend and erodes the abstraction.

If we want to follow the open/closed principle here, I’d prefer we keep `StorageResponse` strict (only the fields we guarantee), and if we genuinely need additional metadata, we should expose it explicitly via an optional standard field that apply across all implementations

That keeps the abstraction clean while still giving us room to evolve.

What are your thoughts?

packages/services/src/interfaces/storageService.ts
```diff
@@ -0,0 +1,42 @@
...
+  uri: string;
+  hash: string;
+} & Extensible;
+
+/**
+ * Configuration for storing a credential in an external storage service
+ */
+export type StorageConfig = {
```
ashleythedeveloper: Suggestion only, keen to get your thoughts.

Rather than using `& Extensible` throughout `StorageConfig`, would it make sense to tighten the contract and push variability into typed generics instead?

For example:

```ts
export type StorageConfig<TParams extends object = {}> = {
  params?: TParams;
  headers?: Record<string, string>;
};
```

and have adapter specific configs like:

```ts
type S3Params = { bucket: string; keyPrefix?: string };
type S3StorageConfig = StorageConfig<S3Params>;
```

This keeps the base config explicit while still allowing each implementation to define exactly what it supports.

Also, should HTTP semantics like `method: 'POST' | 'PUT'` live inside the adapter rather than the interface? If we ever need update semantics, it might be clearer to expose that directly on `IStorageService`:

```ts
store(credential: SignedVerifiableCredential, config?: StoreConfig): Promise<StorageRecord>;
update(documentId: string, credential: SignedVerifiableCredential, config?: StoreConfig): Promise<StorageRecord>;
```

Finally, would it be cleaner to move stable wiring like base URL and default headers into the adapter constructor, leaving `StorageConfig` as per call overrides?

```ts
class HttpStorageService implements IStorageService {
  constructor(
    private readonly baseUrl: string,
    private readonly defaultHeaders: Record<string, string>
  ) {}

  async store(credential: SignedVerifiableCredential, config?: StoreConfig): Promise<StorageRecord> {
    const headers = { ...this.defaultHeaders, ...config?.headers };
    // ...
  }
}
```

packages/services/src/interfaces/storageService.ts
bla-ce: Agree. I also updated the PublishResponse type

packages/services/src/interfaces/storageService.ts
bla-ce: I agree with the above, and that's actually what I did later when implementing the store method. Updating now.

packages/services/src/interfaces/verifiableCredentialService.ts
```diff
@@ -0,0 +1,127 @@
+import type {
```
ashleythedeveloper: Using the same abstraction lens as the IDR and Storage Service discussion, a few things in `IVerifiableCredentialService` are still leaking implementation details and blurring the boundary a bit.

At a spec level, UNTP is aligned with **[W3C VCDM v2](https://www.w3.org/TR/vc-data-model-2.0/)**, but applies an [opinionated subset on top of it](https://untp.unece.org/docs/specification/VerifiableCredentials) (JSON-LD compacted form, JOSE enveloping, specific contexts, status mechanisms, etc). Given that, the service interface should describe the capability in domain terms, not the transport or implementation choices.

### A couple of concrete points

#### 1) `baseURL` and `headers` are transport concerns
Passing these into `sign(...)` makes the interface read like “call a remote signing API” rather than “sign a credential”. IMO, endpoint configuration and auth belong in the constructor or adapter, not the method signature.

#### 2) Return type shape
`SignedVerifiableCredential = { verifiableCredential: W3CVerifiableCredential }` reflects a specific response payload rather than the domain concept. In UNTP, the signed artefact is the **enveloped credential** (JOSE), so that should be the thing the interface returns.

Something along these lines feels cleaner:

```ts
interface IVerifiableCredentialService {
  sign(payload: CredentialPayload): Promise<EnvelopedVerifiableCredential>
  verify(credential: EnvelopedVerifiableCredential): Promise<VerifyResult>
  decode(credential: EnvelopedVerifiableCredential): Promise<DecodedVerifiableCredential>
}
````

### 3) Decision point: VCDM v2 vs UNTP profile types

One thing I think we need to settle is whether the types themselves should model the full [VCDM v2 spec](https://www.w3.org/TR/vc-data-model-2.0/), or the narrower [UNTP profile](https://untp.unece.org/docs/specification/VerifiableCredentials#verifiable-credential-profile).

| Option            | Pros                                        | Cons                                                                                        |
| ----------------- | ------------------------------------------- | ------------------------------------------------------------------------------------------- |
| Full VCDM v2      | Flexible; could accept broader inputs       | Types promise more than the UNTP spec calls for; misrepresents what the RI is demonstrating |
| UNTP profile only | Spec-aligned; honest about what the RI does | Types may need updating if the profile evolves                                              |

My view is that the reference implementation should model the **UNTP profile** in its types. The spec explicitly defines a constrained profile for interoperability reasons, and the RI should demonstrate that profile rather than the full VCDM v2 flexibility. The fact that the underlying implementation (eg VCKit, etc..) can handle broader verification does not mean the RI’s interface should expose that.

What are your thoughts?

**Review by ashleythedeveloper — COMMENTED** [review]
Similar comments as on the Storage Service.

File: packages/services/src/interfaces/identityResolverService.ts [hunk 6]
```diff
@@ -0,0 +1,38 @@
...
+ */
+export type PublishResponse = {
+  enabled: boolean;
+  raw: JSONValue;
+} & Extensible;
+
+/**
+ * Service for publishing credentials to an identity resolver
+ */
+export interface IIdentityResolverService {
```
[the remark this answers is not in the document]
bla-ce: Yes, I agree with the above. I think I misunderstood the purpose of the IDR and simply followed the current implementation in `packages/services/src/linkResolver.service.ts:180`, updating now.

packages/services/src/interfaces/verifiableCredentialService.ts
```diff
@@ -0,0 +1,125 @@
...
+  Extensible,
+  JSONObject,
+  JSONValue,
+  NonEmptyArray,
+  OneOrMany
+} from "@/types";
+
+/** "type" fields can be a string or a non-empty array of strings */
```
ashleythedeveloper: As per [this comment](https://github.com/uncefact/tests-untp/pull/360/changes#r2703386081), I believe we agreed we would constrain the types associated with VCs to the UNTP VCDM profile, not the full v2 VCDM profile.

> One thing I think we need to settle is whether the types themselves should model the full VCDM v2 spec, or the narrower UNTP profile.

> I think the RI should only model the UNTP profile. I think it is better to update the types than to misrepresent what the RI is demonstrating.

ashleythedeveloper (changes requested):

PR #510, opened 2026-04-07
ci: read PLAYGROUND_VERIFICATION_SERVICE_URL from vars context

ashleythedeveloper:
This PR fixes the workflow so the playground deployment actually receives a value for `VERIFICATION_SERVICE_URL`. It was being read as `${{ env.PLAYGROUND_VERIFICATION_SERVICE_URL }}`, which references a workflow `env:` block that doesn't exist anywhere in the file, so the expression silently resolved to an empty string. The Pulumi conditional then dropped it and the deployed container started without `VERIFICATION_SERVICE_URL` set.

Switched both deploy jobs to `${{ vars.PLAYGROUND_VERIFICATION_SERVICE_URL }}` so the GitHub repository/environment variable is read correctly. The token already uses `secrets.` and is unaffected. The infra README has been updated to match.

## Test plan
- [ ] After the next deployment, the playground container has `VERIFICATION_SERVICE_URL` set to the configured value
- [ ] The verify route no longer falls through to "Verification service not configured"


ashleythedeveloper:
@gemini-code-assist this is incorrect. The PR contains two commits:

- `4018c5b` — `.github/workflows/ci_cd-untp-playground.yml` (the actual fix, switching both deploy jobs from `env.` to `vars.`)
- `f7d64c3b` — `packages/untp-playground/infra/README.md` (matching documentation update)

Both files are in the diff. The workflow change is the substantive fix and is present.

